PR #69, opened 2015-01-05
Fixes an issue where pre-providing an answer on a machine with only one platform will throw a RuntimeError.

mitchellallison:
If an answer is given for a machine with only one platform, nothing is popped from the `answers` array. As a result, the final check in `create_some_context` fails with a runtime error.

Using the behaviour of the approach with many platforms will not only remove the choice from the `answers` array, but it will ensure that the platform chosen is indeed the intended choice.


inducer:
Thanks!

PR #74, opened 2015-02-09
Work around in setup.py for use of numpy during build

timrudge:
Currently when installing via pip (pip install pyopencl) build fails because numpy cannot be imported. This seems to be a known issue with the way numpy installation is organised (see http://stackoverflow.com/questions/19919905/how-to-bootstrap-numpy-installation-in-setup-py). The workaround used in this link is to force numpy visibility using "\_\_builtins\_\_.\_\_NUMPY_SETUP\_\_ = False".

I have added this to setup.py, as well as adding numpy to install_requires since it is used when running pyopencl. Tested with "pip install -e <path to modified source>".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,14 @@
 #!/usr/bin/env python
 # -*- coding: latin-1 -*-
+from setuptools.command.build_ext import build_ext as _build_ext
+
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

PR #75, opened 2015-02-10
Revise fix for numpy issue during installation 

timrudge:
Seems to work by just adding numpy to  install_requires and setup_requires.


inducer:
Thanks!

PR #78, opened 2015-06-05
generalized sorting to work on arbitrary kind of scan kernel

rckirby:
This does about what it says -- parameter was added to RadixSort so that one could stuff in a GenericDebugScanKernel (or something else) to change the parallel properties of RadixSort.


inducer:
Thx!

PR #80, opened 2015-07-15
Bitonic sort

inducer:
Get @inferrna's bitonic sort ready for inclusion.

### DIFF
--- a/doc/algorithm.rst
+++ b/doc/algorithm.rst
@@ -271,4 +271,9 @@ Building many variable-size lists
 
 .. autoclass:: ListOfListsBuilder
 
-    .. automethod:: __call__
+Bitonic Sort
+------------
+
+.. module:: pyopencl.bitonic_sort
+
+.. autoclass:: BitonicSort
--- a/pyopencl/algorithm.py
+++ b/pyopencl/algorithm.py
@@ -395,6 +395,7 @@ RADIX_SORT_OUTPUT_STMT_TPL = Template(r"""//CL//
 
 
 # {{{ driver
+
 # import hoisted here to be used as a default argument in the constructor
 from pyopencl.scan import GenericScanKernel
 
@@ -402,6 +403,8 @@ from pyopencl.scan import GenericScanKernel
 class RadixSort(object):
     """Provides a general `radix sort <https://en.wikipedia.org/wiki/Radix_sort>`_
     on the compute device.
+
+    .. seealso:: :class:`pyopencl.algorithm.BitonicSort`
 
     .. versionadded:: 2013.1
     """
--- a/pyopencl/bitonic_sort.py
+++ b/pyopencl/bitonic_sort.py
@@ -1,0 +1,154 @@
+from __future__ import division, with_statement, absolute_import, print_function
+
+__copyright__ = """
+Copyright (c) 2011, Eric Bainville
+Copyright (c) 2015, Ilya Efimoff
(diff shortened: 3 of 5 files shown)


inducer:
On Intel's CPU implementation, this crashes in the tests with OUT_OF_RESOURCES. (likely for overrunning local memory--could this be reduced somehow?)

inducer:
Nvm, fixed. All tests pass on all CL platforms that I test PyOpenCL on.

inferrna:
I'm little worried about speed. Am I right about thinking that @memoize_method does caching all method work results and kernels will not recreated on each call? All other changes looks good for me.

inferrna:
I also reviewed the code and found that this part of code
```python
                        cl.LocalMemory(wg[0]*4*arr.dtype.itemsize),
                        cl.LocalMemory(wg[0]*4*idx.dtype.itemsize),
```
can be safely replaced with
```python
                        cl.LocalMemory(wg[0]*arr.dtype.itemsize),
                        cl.LocalMemory(wg[0]*idx.dtype.itemsize),
```
multiplying to 4 was needed for C4 kernel, that gives very little speedup comparing current version and currently not adapted for multidimensional arrays. I also have added functions that uses c4 into gist https://gist.github.com/inferrna/2202fb8938fbc63e54dc#file-c4-py

inducer:
Yes, `@memoize_method` implements precisely the same kind of instance-level cache that you had written by hand.

I've gone ahead and shrunk the local memory as you indicated.

Thanks for your contribution and the review of the adapted code!

inducer:
Obtw, could this be made to work for float64?

inferrna:
I added 
```
% if dtype=='double':
#pragma OPENCL EXTENSION cl_khr_fp64 : enable
% else:
#pragma OPENCL EXTENSION cl_khr_fp64 : disable
% endif
```
into defines template and got float64 working with pocl. Test on beignet was failed because of not full 1.2 support - it even cann't generate float64 array with pyopencl (error: use of type 'double' requires cl_khr_fp64 extension to be enabled). On weekend I'll try run code on AMD GPU's and report result.

inducer:
Looks like double works on AMD and Intel. It just breaks on my POCL (I should report that.). I've xfailed the double tests on POCL for now.

inferrna:
I found an (possible) bug and also unused typedef in bitonic_sort_templates.py. First, string
```C
typedef ${idxtype}2 idx_t2;
```
can be safely removed from bitonic_sort_templates.py.

Second:
The bug is in function dtype_to_ctype(): it converts uint32 as 'unsigned' but there isn't described unsignedn vector type at https://www.khronos.org/registry/cl/sdk/1.2/docs/man/xhtml/vectorDataTypes.html . And also dtype_to_ctype() converts other integer types non vector-friendly.
It will break any template expressions like ${dtype}n. In my case beignet fails to compile kernel with
```C
typedef unsigned2 idx_t2;
```
as it turned out, I have planned, but never used idx_t2 in this template.

inducer:
I've gone ahead and removed the typedef in a177ba4.

inducer:
As of 293ee46, PyOpenCL should use more sensible type names, as you suggest. That said, the method you suggest for constructing vector type names is brittle and should not be used. Instead, PyOpenCL's array facility already provides a way of predictably making vector type names:

https://github.com/pyopencl/pyopencl/blob/master/pyopencl/array.py#L149

PR #82, opened 2015-08-03
fix shebang on configure.py

larsmans:
6e24507b91f5f27c8f0443ba3d6ff1e307d30a87 broke the shebang on `configure.py`. I must admit that the error message for this was quite asuming:

```
from: can't read /var/mail/__future__
```

... but I decided to fix it anyway.


inducer:
Thanks for your contribution!

PR #83, opened 2015-08-06
Fix build on Ubuntu/GCC 4.8.4 and 4.9.0

### LINKED ISSUE #81: Build error on dev version
I'm able to install the newest release (2015.1) using the instructions [here](http://wiki.tiker.net/PyOpenCL/Installation/Linux/Ubuntu). (By the way, these instructions need to be updated, since `configure.py` no longer seems to take boost arguments. [EDIT: Sorry, it's just in `master` that boost arguments no longer work. They're fine in `v2015.1`.] Also, it's a bit confusing having documentation both at that site and at http://documen.tician.de/pyopencl/.)

But when I try to build what's currently in master, I get a bunch of errors. I've posted the first one below.
```
x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -fwrapv -Wall -O3 -DNDEBUG -fPIC -DPYGPU_PACKAGE=pyopencl -DPYGPU_PYOPENCL=1 -DPYOPENCL_USE_DEVICE_FISSION=1 -I/opt/AMDAPP/include -Isrc/c_wrapper/ -I/usr/include/python2.7 -c src/c_wrapper/wrap_cl.cpp -o build/temp.linux-x86_64-2.7/src/c_wrapper/wrap_cl.o -std=c++0x
In file included from src/c_wrapper/clhelper.h:1:0,
                 from src/c_wrapper/wrap_cl.cpp:2:
src/c_wrapper/error.h: In instantiation of ‘void call_guarded(cl_int (*)(ArgTypes ...), const char*, ArgTypes2&& ...) [with ArgTypes2 = {device* const, int, ArgBuffer<_cl_platform_id*, (ArgType)1>, std::nullptr_t}; ArgTypes = {_cl_device_id*, unsigned int, long unsigned int, void*, long unsigned int*}; cl_int = int]’:
src/c_wrapper/device.h:39:17:   required from here
src/c_wrapper/error.h:218:51: error: no matching function for call to ‘CLArgPack<device* const, int, ArgBuffer<_cl_platform_id*, (ArgType)1>, std::nullptr_t>::clcall(cl_int (*&)(_cl_device_id*, unsigned int, long unsigned int, void*, long unsigned int*), const char*&)’
     cl_int status_code = argpack.clcall(func, name);
                                                   ^
```

larsmans:
The error was that `clobj<CLType>` was passed where `CLType` was expected. I'm not sure why this was ever supposed to work; I tried putting a type-cast operator on `clobj`, but that didn't work.

Fixes #81.

### DIFF
--- a/src/c_wrapper/buffer.cpp
+++ b/src/c_wrapper/buffer.cpp
@@ -18,7 +18,7 @@ buffer::get_sub_region(size_t orig, size_t size, cl_mem_flags flags) const
     cl_buffer_region reg = {orig, size};
 
     auto mem = retry_mem_error([&] {
-            return pyopencl_call_guarded(clCreateSubBuffer, this, flags,
+            return pyopencl_call_guarded(clCreateSubBuffer, data(), flags,
                                          CL_BUFFER_CREATE_TYPE_REGION, &reg);
         });
     return new_buffer(mem);
@@ -28,7 +28,7 @@ PYOPENCL_USE_RESULT buffer*
 buffer::getitem(ssize_t start, ssize_t end) const
 {
     ssize_t length;
-    pyopencl_call_guarded(clGetMemObjectInfo, this, CL_MEM_SIZE,
+    pyopencl_call_guarded(clGetMemObjectInfo, data(), CL_MEM_SIZE,
                           size_arg(length), nullptr);
     if (PYOPENCL_UNLIKELY(length <= 0))
         throw clerror("Buffer.__getitem__", CL_INVALID_VALUE,
@@ -45,7 +45,7 @@ buffer::getitem(ssize_t start, ssize_t end) const
         throw clerror("Buffer.__getitem__", CL_INVALID_VALUE,
                       "Buffer slice should have end > start >= 0");
     cl_mem_flags flags;
-    pyopencl_call_guarded(clGetMemObjectInfo, this, CL_MEM_FLAGS,
+    pyopencl_call_guarded(clGetMemObjectInfo, data(), CL_MEM_FLAGS,
                           size_arg(flags), nullptr);
     flags &= ~CL_MEM_COPY_HOST_PTR;
     return get_sub_region((size_t)start, (size_t)(end - start), flags);
--- a/src/c_wrapper/command_queue.cpp
+++ b/src/c_wrapper/command_queue.cpp
@@ -13,7 +13,7 @@ template void print_buf<cl_command_queue>(
 
 command_queue::~command_queue()
 {
-    pyopencl_call_guarded_cleanup(clReleaseCommandQueue, this);
+    pyopencl_call_guarded_cleanup(clReleaseCommandQueue, data());
 }
 
 generic_info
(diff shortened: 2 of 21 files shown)


yuyichao:
`this` is used instead of `data()` in order to trace which object is holding the data. If you have `DEBUG_ON` (either at compile time or at runtime) you should see that each call to the OpenCL API are printed with the corresponding PyOpenCL c++ wrapper address as well as the OpenCL object address. Using `data()` will make it only print the OpenCL pointer and at some point the ability to track c++ object pointer was important for debugging.

The general procedure is
1. On creation of an `CLArgPack` (which holds all the arguments), arguments is converted to `CLArg`.
2. The `CLArg` is templated on the type and in particular a specialization is defined in `clobj.h` to handle using `clobj` as argument.
3. When calling the function, `CLArg::convert` is called to retrive the real argument to the function and for `ClArg<clobj>` this should return the corresponding OpenCL object pointer.
4. As mentioned above, the convert is done at call time so that the type information is preserved for debug printing.

Can you check the output of the compiler to figure out what exactly it is complaining? (it should tell you why it cannot match each template)

larsmans:
Here's the compiler output for the first file:

https://gist.github.com/larsmans/64e4ebc5e2ed36d548e7

There's megabytes more of this, but it's all the same kind of error.

Starkstroem:
I still got the same error as above with 14.04 and gcc 4.9.
Anyone able to install pyopencl successfully?

inducer:
@Starkstroem It builds without a problem for me on gcc 5.2. If you keep having issues, try [2015.1](https://pypi.python.org/pypi/pyopencl/2015.1). It has the same interface on the Python side as the newer code, and has been around longer. Nonetheless, gcc 4.9 is not so old that we shouldn't at least *try* to get it to work.

inducer:
@yuyichao Any clues what's happening here?

yuyichao:
I couldn't reproduce this on GCC 4.8 or 4.9 so I'm not really sure what's happening.

[As I said](https://github.com/pyopencl/pyopencl/pull/83#issuecomment-128425445), this automatic casting was only useful for debugging (to track pointers) and it is certainly ok to avoid it if it's giving people to much issues.

inducer:
e12f38a38729dab3e6ac36749d25fc0bce4dbfd7 is a version of this that retains debug info if it's being built on gcc >= 5.2 and otherwise is as compatible as possible. It builds OK for me on Ubuntu 14.04. Please let me know if this resolves the issue.

inducer:
Also, thanks @larsmans for the initial patch, and sorry for the long delay in looking at this.

UniqueFool:
thanks for fixing this - however, would someone please so kind to update the pip tarball, so that this fix actually propagates (currently, "pip install pyopencl" will fail on all ubuntu/gcc systems mentioned above)

Referring to: https://pypi.python.org/pypi/pyopencl

Thanks

inducer:
I've just upload 2015.2.2 with these fixes built in.

UniqueFool:
thanks for being so responsive - unfortunately, it seems, there are still some issues on:

No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.3 LTS
Release:	14.04
Codename:	trusty
gcc (Ubuntu 4.8.4-2ubuntu1~14.04) 4.8.4


See: http://pastebin.com/mMzKrKS1

inducer:
Ugh, sorry. That one slipped through. Should be fixed in 4be4cf5. Can you try current git?

UniqueFool:
I can confirm that building from git, everything is building correctly - thanks again for being so responsive, WTG !
Once you have updated the pip package, I will try again and report back here.

thanks

inducer:
2015.2.3 is on the package index.

UniqueFool:
success, terrific maintenance job :+1:

PR #86, opened 2015-08-07
Use sscanf instead of std::regex

### LINKED ISSUE #85: Import crashes due to a regex_error
This is on the branch submitted as #83, but I didn't touch regexps at all so I assume it will hold for a different fix of the build problem as well:

```
$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyopencl as cl
terminate called after throwing an instance of 'std::regex_error'
  what():  regex_error
Aborted
```

yuyichao:
Fix #85

### DIFF
--- a/src/c_wrapper/platform.cpp
+++ b/src/c_wrapper/platform.cpp
@@ -3,7 +3,6 @@
 #include "clhelper.h"
 
 #include <stdlib.h>
-#include <regex>
 
 template class clobj<cl_platform_id>;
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

larsmans:
Thanks indeed!

PR #106, opened 2015-12-22
Fixes to enqueue_copy_buffer_rect

### LINKED ISSUE #105: rectangle copy with buffer pitches broken in 2015.2.3
In `pyopencl/cffi_cl.py` function `_enqueue_copy_buffer_rect`, the variables `src_pitches_l` and `dst_pitches_l` are only assigned if the corresponding pitches array is given as None. This causes use-before-definition errors if pitches are actually passed.

bmerry:
There are two bugs:
1. If src_pitches or dst_pitches is specified, there is an uninitialised
   variable.
2. If ConstBuffer is given a short array and content==0, the internal
   buffer is not zero-filled, causing garbage to be passed on.

The latter probably affects a lot of other functions, but hopefully in a
positive way.

Fixes #105.


inducer:
Looks good. Thanks for the fix!

PR #107, opened 2016-01-14
Fix indexing with reverse slices

SyamGadde:
An earlier commit fixed a rounding issue for slices with steps > 1, but it broke reverse indexing, where step < 0.  This fix should re-enable reverse indexing.


inducer:
Thanks for your contribution!

PR #115, opened 2016-03-02
Reworked how PyOpenCL module finds its path

jmc734:
[Previously](https://github.com/pyopencl/pyopencl/blob/master/pyopencl/__init__.py#L234), `pkg_resources` was used to find the location of the PyOpenCL module so that the include path could be constructed.
This commit replaces that by simply using the [`__file__`](https://github.com/jmc734/pyopencl/blob/master/pyopencl/__init__.py#L237) and [`os.path`](https://github.com/jmc734/pyopencl/blob/master/pyopencl/__init__.py#L235) utilities

This change fixes an incompatibility with PyOpenCL and pyinstaller/cxfreeze/etc. When frozen, distutils is not able to resolve the PyOpenCL package directly and subsequently raises a DistributionNotFound error. Since we are the PyOpenCL package, we don't need to be using distutils to resolve it. This issue was [reported](https://lists.tiker.net/pipermail/pyopencl/2014-October/001832.html) by someone else in the pyopencl mailing list as well.

### DIFF
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -232,13 +232,21 @@ del _size_t_char
 # {{{ find pyopencl shipped source code
 
 def _find_pyopencl_include_path():
-    from os.path import join, abspath, dirname, exists
+    from pkg_resources import Requirement, resource_filename, DistributionNotFound
+    try:
+        # Try to find the resource with pkg_resources (the recommended setuptools approach)
+        return resource_filename(Requirement.parse("pyopencl2"), "pyopencl/cl")
+    except DistributionNotFound:
+        # If pkg_resources can't find it (e.g. if the module is part of a frozen application),
+        # try to find the include path in the same directory as this file 
+        from os.path import join, abspath, dirname, exists
     
-    include_path = join(abspath(dirname(__file__)), "cl")
-    if not exists(include_path):
-        raise RuntimeError("Could not located PyOpenCL include path")
-        
-    return include_path
+        include_path = join(abspath(dirname(__file__)), "cl")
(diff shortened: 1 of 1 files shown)


jmc734:
If there is some reason that I am not aware of for using `pkg_resources`, it may still be worthwhile to include this as a fallback.

inducer:
https://github.com/pyopencl/pyopencl/commit/923587036c75cccb6764cabd368ddca9d29521ad is the change that went from something like what you propose to the current `pkg_resources` version. As a confounding factor, the same commit also switches from setuptools' `data_files` to `package_data`. The `data_files` was, as you can see, a mess. The selling point of `pkg_resources` is (IIRC) that it's compatible with "wheels" and "eggs", and setuptools advertises it as "the correct way" to access package data. So I think I would like to retain it as a fallback.

jmc734:
After further [research](http://peak.telecommunity.com/DevCenter/PythonEggs#accessing-package-resources), `pkg_resources` does seem to be the preferred approach though there is still the issue with frozen applications. In addition to the `__file__` method described above, another common approach is using the `imp` module but as of mid-2013 it seems to be [deprecated](http://bugs.python.org/issue17177). For that reason, I chose to have the code [first](https://github.com/jmc734/pyopencl/blob/8a55b19ff134592fcdaacd619c3211e61743a822/pyopencl/__init__.py#L236) attempt to use `pkg_resources` and [fallback](https://github.com/jmc734/pyopencl/blob/8a55b19ff134592fcdaacd619c3211e61743a822/pyopencl/__init__.py#L239) to using `os.path` and `__file__`.

inducer:
That looks good. Thanks for the contribution!

PR #120, opened 2016-03-30
Allow for spaces in installation path

AlexRothwell7:
Fixes INVALID_COMPUTER_OPTIONS error if PyOpenCL installation path includes spaces

### DIFF
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -257,7 +257,7 @@ def _find_pyopencl_include_path():
 # {{{ Program (including caching support)
 
 _DEFAULT_BUILD_OPTIONS = []
-_DEFAULT_INCLUDE_OPTIONS = ["-I", _find_pyopencl_include_path()]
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for your contribution. 9b02b38887aa9a6851bcf34853ff8a07bbc638d5 (and subsequent fixes) does a more careful job quoting build options. That should resolve your issue. Let me know if that's not the case.

AlexRothwell7:
Thanks. The fixes didn't seem to quite fix the problem as it seems that to quote the options on Windows you need to use double quotes rather than single quotes. Changing:
`return b"'" + s.replace(b"'", b"'\"'\"'") + b"'"`
to
`return b'"' + s.replace(b"'", b"'\"'\"'") + b'"'`
seems to have fixed the problem here

inducer:
Attempted fix in c85ada5. Please test. Thx!

AlexRothwell7:
Looks good. Just need to change the assert to
`assert b'"' not in s`
to avoid comparing bytes and strings and then it works. Thanks!

inducer:
Done. Thanks for the report!

PR #124, opened 2016-04-22
Adding ability to build with GL interop from pip command

ernwa:
This is my first pull request, so sorry if it's not right.

The old method of pip install defaults to not including opengl interop as some setups will not have GL. A build switch is used with configure.py to turn it on, but that's not something that could be done from setup.py. I have modified setup.py such that it can take extra arguments, which are able to be passed in from pip using the "--install-option" switch. This seems to allow me to use pip to turn on gl interop in the build without any special configuration files or commands. I hope it will work similarly from PyPI.

tested in the "pyopencl" directory using
```pip install . --install-option="CL_ENABLE_GL"```
followed by running the gl_interop_demo.py from the examples directory.

### DIFF
--- a/pyopencl/version.py
+++ b/pyopencl/version.py
@@ -1,3 +1,3 @@
-VERSION = (2015, 2, 4)
+VERSION = (2016, 1)
 VERSION_STATUS = ""
 VERSION_TEXT = ".".join(str(x) for x in VERSION) + VERSION_STATUS
--- a/setup.py
+++ b/setup.py
@@ -96,6 +96,19 @@ def main():
     conf = get_config(get_config_schema(),
             warn_about_no_config=False)
 
+
+    custom_args = ['CL_ENABLE_GL', 'CL_TRACE', 'CL_PRETEND_VERSION']
+
+    new_argv = []
+    for word in sys.argv:
+        arg = word.split('=', 1)
+        k, v = (arg[0], True) if (len(arg) == 1) else arg
+        if k in custom_args:    # if it's one of our options
+            conf[k] = v         # update config
+        else:
+            new_argv.append(word)     # remove from argv so setuptools doesn't get mad
+    sys.argv = new_argv
+
     extra_defines = {}
 
     extra_defines["PYGPU_PACKAGE"] = "pyopencl"

### CODE REVIEW

**Comment by inducer** [comment]
Thanks for your contribution. I've decided to implement something like this in aksetup instead (so that it's available in PyCUDA and all my other packages, too), in 24e54acc52e723669e38e10a1aed1899bf1add9.

Now `pip install . --install-option="--conf:CL_ENABLE_GL=1` should do the trick.

**Comment by ernwa** [comment]
sounds good! Will this be rolled out to pypi soon?

**Comment by inducer** [comment]
I could do that, although I'm not in a particular rush. Do you have a particular target date in mind?

**Comment by ernwa** [comment]
It just seems to be a major enhancement from the usability standpoint, even though perhaps under the hood not much is changed. If you just released 2016.1 you're at a good point to do a minor release allowing easy gl interop and then can leave it for awhile. 
My personal reason is that our lab has a bunch of people using pyopencl with OpenGL and it makes deploying to servers much much quicker than going through a git clone process. It's nice to be able to just pip install whatever and get on with things. 
Really it's up to you but sooner would be nicer :-)

Sent from Outlook Mobile




On Tue, Apr 26, 2016 at 11:07 AM -0700, "Andreas Klöckner" <notifications@github.com> wrote:












I could do that, although I'm not in a particular rush. Do you have a particular target date in mind?

**Comment by inducer** [comment]
Thanks for the feedback--it didn't happen, life got busy. And by now I've got a number of bigger changes queued up that aren't quite fully baked (SVM support, new random number generators).

In the meantime, you can do
```
pip install git+https://github.com/pyopencl/pyopencl.git
```
(plus whatever options) to install directly from git.

ernwa:
sounds like a start....have you considered somehow separating the core
opencl functionality from higher-level constructs like random numbers and
svm support, or maintainance vs feature releases? Anyway, thank you for
fixing the issue and i hope it makes it to an official release soon.

inducer:
The thought has crossed my mind. This'll happen at the latest once the PyOpenCL array machinery gets replaced by code generated by [something more capable](/inducer/loopy).

PR #126, opened 2016-05-31
Fixes  for _enqueue_*_buffer_rect

Shane-J-Latham:
Hi, here are a couple of fixes along the lines of commit 08bc620955224822ce79476e78904a4c58da24ae (#105) to fix *unitialised variable* error in `pyopencl.cffi_cl._enqueue_read_buffer_rect` and `pyopencl.cffi_cl._enqueue_write_buffer_rect` when the `buffer_pitches` and `host_pitches` arguments are specified (i.e. not `None`). Also added `test/test_enqueue_copy.py` unit-test (which uses `unittest` rather than `pytest`).

Have also *tuple-ised* the pitches arguments in `pyopencl.cffi_cl._enqueue_copy_buffer_rect`.

### DIFF
--- a/pyopencl/cffi_cl.py
+++ b/pyopencl/cffi_cl.py
@@ -1358,9 +1358,16 @@ def _enqueue_read_buffer_rect(queue, mem, hostbuf, buffer_origin,
     if buffer_pitches is None:
         buffer_pitches = _ffi.NULL
         buffer_pitches_l = 0
+    else:
+        buffer_pitches = tuple(buffer_pitches)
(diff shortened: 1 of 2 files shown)


inducer:
Looks good. Thanks for the contribution!

inducer:
FYI: c7bbe640d7c034005c6d5cb59522104bdbc08970

Shane-J-Latham:
Hah hah, here I was hoping you would convert the `pytest`'s to `unittest` 8-)...I probably
could have made it `pytest` in the first place easy enough...thanks.

PR #128, opened 2016-06-09
Wrappers for nvidia CL_DEVICE_PCI_BUS_ID_NV, etc device attributes

Shane-J-Latham:
Hi, sorry to bug you again...Wondering if it is OK to add some nvidia-specific
device attributes:

   a. `CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV`
   b. `CL_DEVICE_PCI_BUS_ID_NV`
   c. `CL_DEVICE_PCI_SLOT_ID_NV`

Trying to do `multiprocessing` on multiple GPUs and the PCI_BUS_ID and PCI_SLOT_ID
attributes seem to be the only/best way to identify a device? `ATTRIBUTE_ASYNC_ENGINE_COUNT` is there just for completeness.

Not sure if the additions in `src/c_wrapper/wrap_cl.h` are in the right place,
seems to be the only spot where `cl_ext.h` is included...

I would have had a crack at the AMD *topology* but I don't have any AMD hardware
to test it on....

### DIFF
--- a/src/c_wrapper/wrap_cl.h
+++ b/src/c_wrapper/wrap_cl.h
@@ -157,6 +157,20 @@ extern "C" {
 #define PYOPENCL_USE_RESULT
 #endif
 
+#ifdef CL_DEVICE_COMPUTE_CAPABILITY_MAJOR_NV
+
+#ifndef CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV
+#define CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV   0x4007
+#endif
+#ifndef CL_DEVICE_PCI_BUS_ID_NV
+#define CL_DEVICE_PCI_BUS_ID_NV                     0x4008
+#endif
+#ifndef CL_DEVICE_PCI_SLOT_ID_NV
+#define CL_DEVICE_PCI_SLOT_ID_NV                    0x4009
+#endif
+
+#endif // CL_DEVICE_COMPUTE_CAPABILITY_MAJOR_NV
+
 #endif
 
 // vim: foldmethod=marker
--- a/doc/make_constants.py
+++ b/doc/make_constants.py
@@ -97,6 +97,9 @@ const_ext_lookup = {
             "GPU_OVERLAP_NV": nv_devattr,
             "KERNEL_EXEC_TIMEOUT_NV": nv_devattr,
             "INTEGRATED_MEMORY_NV": nv_devattr,
+            "ATTRIBUTE_ASYNC_ENGINE_COUNT_NV": nv_devattr,
+            "PCI_BUS_ID_NV": nv_devattr,
+            "PCI_SLOT_ID_NV": nv_devattr,
 
             "DOUBLE_FP_CONFIG":
             ("cl_khr_fp64", "2011.1"),
--- a/src/c_wrapper/device.cpp
+++ b/src/c_wrapper/device.cpp
@@ -143,6 +143,9 @@ device::get_info(cl_uint param_name) const
     case CL_DEVICE_COMPUTE_CAPABILITY_MINOR_NV:
     case CL_DEVICE_REGISTERS_PER_BLOCK_NV:
     case CL_DEVICE_WARP_SIZE_NV:
+    case CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV:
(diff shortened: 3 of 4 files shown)


src/c_wrapper/wrap_cl.h
```diff
@@ -157,6 +157,20 @@ extern "C" {
...
+
+#ifndef CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV
+#define CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV   0x4007
+#endif
+#ifndef CL_DEVICE_PCI_BUS_ID_NV
+#define CL_DEVICE_PCI_BUS_ID_NV                     0x4008
+#endif
+#ifndef CL_DEVICE_PCI_SLOT_ID_NV
+#define CL_DEVICE_PCI_SLOT_ID_NV                    0x4009
+#endif
```
inducer: I'm less fond of carrying around our own definitions of these constants--why can't those come from `cl_ext.h`?

Shane-J-Latham:
> I'm less fond of carrying around our own definitions of these constants

Yeah, I can understand why you are not fond of this...

My thinking is that if pyopencl doesn't `#define` them, on systems with multiple open-cl
implementations you have to make sure that pyopencl is compiled against the nvidia opencl
implementation (so that the nvidia `CL/cl_ext.h` is used to get the `#defines`) in order to use
these device attributes.

Maybe borrow [ext.h](https://github.com/Oblomov/clinfo/blob/master/src/ext.h) from [clinfo](https://github.com/Oblomov/clinfo) ?

inducer:
Why not make it the user's problem to have an up-to-date `cl_ext.h`?

inducer:
Oh, sorry, didn't fully read what you wrote. IMO, CL implementations installing `libOpenCL.so` and headers is just a terrible idea, and unsustainable for the reason you give. Everyone should just use [ocl-icd](https://github.com/OCL-dev/ocl-icd). I'm hesitant to ship `ext.h`, too--what if that falls out of date?

That said, I see your practical problem, and I could be talked into shipping it and optionally using it. Conceivably even on-by-default, although that's pushing it.

Shane-J-Latham:
> Everyone should just use ocl-icd.

I am not going to disagree with this philosophical point of view, non-standard vendor extensions make life a bit tougher. But  it doesn't look like [ocl-icd](https://github.com/OCL-dev/ocl-icd) is going to help me get the PCI bus id?

> I could be talked into shipping it and optionally using it.

Hooray, how could I talk you into it?

By "optionally", do you mean a build option:
```
python setup.py build --with-clinfo-ext
```
?

> Conceivably even on-by-default, although that's pushing it.

I am not going to pretend I can envisage the consequences of having it on by default (your call). I can see the value in being able to turn it on/off though.

inducer:
> I am not going to disagree with this philosophical point of view, non-standard vendor extensions make life a bit tougher. But it doesn't look like ocl-icd is going to help me get the PCI bus id?

Aww crap. I thought they had the good sense of just at least emailing someone at Khronos. But then, what the hell do I know.

> By "optionally", do you mean a build option:

 Sadly, I've made my own widgetry, before I knew better:
https://github.com/pyopencl/pyopencl/blob/master/setup.py#L36

> I am not going to pretend I can envisage the consequences of having it on by default (your call). I can see the value in being able to turn it on/off though.

Let's try it. If it blows up epically, we can always point at the switch while quickly backpedaling.

Shane-J-Latham:
> Sadly, I've made my own widgetry, before I knew better:
> https://github.com/pyopencl/pyopencl/blob/master/setup.py#L36

OK, looks like I can add a `Switch("CL_PRETEND_EXT",...)` in there....

`#include clinfo_ext.h` instead of `#include CL/cl_ext.h` in `src/c_wrapper/wrap_cl.h`.

Change the export macros a bit to guard the attribute exports (in `src/c_wrapper/wrap_constants.cpp` and `src/c_wrapper/device.cpp`) with `#ifdef CL_DEVICE_<ATTR>`?


Want me to give it a try?

inducer:
Yep. :)

inducer:
Maybe CL_USE_SHIPPED_EXT?

Shane-J-Latham:
> Maybe CL_USE_SHIPPED_EXT?

OK.

Shane-J-Latham:
Closing, see #129.

PR #129, opened 2016-06-13
Ship (and optionally use) ext.h from clinfo to replace CL/cl_ext.h

Shane-J-Latham:
OK, here is more general form of #128, replacing use of `CL/cl_ext.h` with `src/c_wrapper/clinfo_ext.h` ([ext.h](https://github.com/Oblomov/clinfo/blob/master/src/ext.h) from [clinfo](https://github.com/Oblomov/clinfo)).

a. Added `CL_USE_SHIPPED_EXT` switch in `setup.py`
b. Added file `src/c_wrapper/clinfo_ext.h` which is used as `#include` replacement for `CL/cl_ext.h` and includes vendor specific definitions like `CL_DEVICE_PCI_BUS_ID_NV` which are not defined in the Khronos headers.
c. Attempt at python export of the `cl_device_topology_amd` struct (similar to the way `cl_image_format` has been exposed). I haven't tested::
```
   topology = pyopencl.Device.get_info(pyopencl.device_info.TOPOLOGY_AMD)
```
Likewise for the intel-me, intel-ss, qcom and altera attributes, not sure if I should have bothered with these?

### DIFF
--- a/cl_types.h
+++ b/cl_types.h
@@ -105,6 +105,12 @@ typedef struct _cl_buffer_region {
 
 /* cl_ext.h */
 
+typedef union
+{
+    struct { cl_uint type; cl_uint data[5]; } raw;
+    struct { cl_uint type; cl_char unused[17]; cl_char bus; cl_char device; cl_char function; } pcie;
+} cl_device_topology_amd;
+
 /*
 typedef cl_ulong  cl_device_partition_property_ext;
 typedef cl_uint   cl_image_pitch_info_qcom;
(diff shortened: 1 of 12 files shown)


inducer:
FYI: Once https://gitlab.tiker.net/inducer/pyopencl/merge_requests/36 goes in, this will no longer be the default. It caused this (intelligible) error message:

```
pyopencl.cffi_cl.LogicError: clGetPlatformIDs failed: PLATFORM_NOT_FOUND_KHR
```

to become this:

```
pyopencl.cffi_cl.LogicError: clGetPlatformIDs failed: <unknown error -1001>
```

Since that's likely the first thing people see of PyOpenCL, it is important to me that that error comes out reasonable.

PR #138, opened 2016-09-07
Fix error when decoding a NULL string

jcarrano-eiwa:
If some device info is NULL  then this line would cause a RuntimeError: cannot use string() on <cdata 'char *' NULL>.

For example, in my PC, the Intel CPU runtime under Linux has BUILT_IN_KERNELS = NULL.


inducer:
Thanks!

Issue #130, opened 2016-07-05
RuntimeError: cannot discover struct layout 


embg:
I'm trying to use a struct type with pyopencl as described [here](https://documen.tician.de/pyopencl/howto.html). However, I get an error:
```
Traceback (most recent call last):
  File "/home/emg3/miniconda2/envs/pysr/lib/python3.5/site-packages/pytools/__init__.py", line 427, in _deco
    return func._memoize_dic[args]
AttributeError: 'function' object has no attribute '_memoize_dic'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "mwe.py", line 10, in <module>
    my_struct, my_struct_c_decl = cl.tools.match_dtype_to_c_struct(ctx.devices[0], "my_struct", my_struct)
  File "<decorator-gen-3>", line 2, in match_dtype_to_c_struct
  File "/home/emg3/miniconda2/envs/pysr/lib/python3.5/site-packages/pytools/__init__.py", line 430, in _deco
    result = func(*args)
  File "/home/emg3/miniconda2/envs/pysr/lib/python3.5/site-packages/pyopencl/tools.py", line 605, in match_dtype_to_c_struct
    "cannot discover struct layout on '%s'" % device)
RuntimeError: cannot discover struct layout on '<pyopencl.Device 'Quadro NVS 295' on 'NVIDIA CUDA' at 0x24cc120>'
```
Here's a MWE:
```
import pyopencl as cl
import pyopencl.tools
import numpy as np

ctx = cl.create_some_context(interactive=False)
my_struct = np.dtype([('val', np.float32),
                      ('op',  np.uint8),
                      ('a',   np.uint8),
                      ('b',   np.uint8)])
my_struct, my_struct_c_decl = cl.tools.match_dtype_to_c_struct(ctx.devices[0], "my_struct", my_struct)
#RuntimeError: cannot discover struct layout on '<pyopencl.Device 'Quadro NVS 295' on 'NVIDIA CUDA' at 0x1917670>'
```
I found that it works fine if I replace uint8 with uint32, but the error persists with uint16.

inducer:
Doesn't reproduce for me on any of these. What version of PyOpenCL? What version of the Nvidia driver?

```
[0] <pyopencl.Platform 'NVIDIA CUDA' at 0x27b6310>
[1] <pyopencl.Platform 'AMD Accelerated Parallel Processing' at 0x7f7dcf6b5430>
[...]

PR #139, opened 2016-09-09
fix issue #130

Poulpy21:
Fix the match_dtype_to_c_struct function in the case where numpy.intp size differs from sizeof(size_t) in opencl. Most of the time this will happen on a 64bit system with an opencl device that declares CL_DEVICE_ADDRESS_BITS to only 32bits.


inducer:
Great catch! Thanks for the patch.

PR #140, opened 2016-09-12
Better checking for stride equality

hunse:
Ignore dimensions where the shape is 1, since the stride in these
dimensions does not matter. This is necessary when the Numpy flag
``NPY_RELAXED_STRIDES_CHECKING`` is set.

Depends on inducer/compyte#25.
Second part of the fix for pyopencl/pyopencl#137.


inducer:
Looks great. Thanks!

PR #141, opened 2016-09-15
Update benchmark.py - Fixing compiler warning

arashTaher:
Fixing CompilerWarning: `<program source>: warning: double precision constant requires cl_khr_fp64, casting to single precision`
Based on [documentation](https://www.khronos.org/registry/cl/sdk/1.0/docs/man/xhtml/cl_khr_fp64.html), An application that wants to use double will need to include the directive before any double precision data type is declared in the kernel code.

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -54,6 +54,7 @@ for platform in cl.get_platforms():
         dest_buf = cl.Buffer(ctx, mf.WRITE_ONLY, b.nbytes)
 
         prg = cl.Program(ctx, """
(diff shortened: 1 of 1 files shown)


inducer:
Since this kernel is all floats, the right way to fix that would be to turn that `2.0` into `2.0f`.

arashTaher:
@inducer Yep, you're right. Sorry, newbie here. I changed it.

inducer:
Looks good now, thanks for the patch!

PR #146, opened 2016-10-08
Image processing median filter

ShahzaibGill:
Example showing how to leverage OpenCL to do image processing. The kernel function applies a median filter on each pixel of a gray-scale image. This example may be useful for anyone using OpenCL to do image processing in parallel (pixel by pixel).


inducer:
Nice! Thanks for submitting this.

PR #151, opened 2016-11-03
Fix error messages and ability to pass a queue in all cases of setitem flow

hrfuller:
I fixed some confusing exception and warning messages. I also noticed that contrary to the docstring for `setitem`, it was not possible to pass a queue as an arg into `mutli_put` from `setitem`.


inducer:
Thanks for the patch, even if it turned out that all this was already done. :)

hrfuller:
Some of this is still outstanding, but I realized I was working off an old fork. I'm going to submit another pull request that is current.

PR #152, opened 2016-11-03
Fix Exception message and allow queue arg usage for `multi_put`.

hrfuller:
I noticed a confusing Exception message missing a `not`. Also fix issue where the passed queue arg in `Array.setitem` was not being passed onto the `multi_put` call.


inducer:
Thanks!

PR #157, opened 2016-11-25
Fix typo in package name while fetching resource filename: s/pyopencl…

janbraiins:
Hi,

I believe the fix below is a typo in the code where a resource references pyopencl2. Unfortunately, this prevents PyInstaller from bundling the library properly when e.g. the following hook is being used:

```
# We need metadata for pyopencl
from PyInstaller.utils.hooks import copy_metadata
datas = copy_metadata('pyopencl')
```

### DIFF
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -194,7 +194,7 @@ def _find_pyopencl_include_path():
     try:
         # Try to find the resource with pkg_resources (the recommended
         # setuptools approach)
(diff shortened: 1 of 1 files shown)


inducer:
Good catch. Also: weird--where did that come from?

At any rate, thanks for the patch!

janbraiins:
I've looked into git annotations and found the commit where this appeared. Anyway, thanks for merging this quickly.

PR #167, opened 2017-01-18
Apple AMD Radeon GPU bug-fix

hightower8083:
The fix in SCAN_INTERVALS_SOURCE source is introduced, which fixes a compiler crash on AMD Radeon GPU on Apple platform.
For this structed data pointers, ldata[:][:].value, are
replaced by temporal variables tmp and tmp_aux before passing to SCAN_EXPR.
To avoid leftover-worning tmp_aux was added to_PREFIX_WORDS.

### DIFF
--- a/pyopencl/scan.py
+++ b/pyopencl/scan.py
@@ -295,8 +295,11 @@ void ${kernel_name}(
 
             if (LID_0 == 0 && unit_base != interval_begin)
             {
+                scan_type tmp = ldata[K][WG_SIZE - 1].value;
+                scan_type tmp_aux = ldata[0][0].value;
+
                 ldata[0][0].value = SCAN_EXPR(
-                    ldata[K][WG_SIZE - 1].value, ldata[0][0].value,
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for your contribution!

inducer:
Thanks again for your work on this. Could you perhaps you also go through the tests and remove any Apple `xfail`s that have now become redundant? Sorry that I didn't think of that during the review.

hightower8083:
Of all tests this affects only ```test_algorithm.py```, where the GenericScanKernel is used. In my case, I don't have the ```xfail``` related to this bug. If I run the test with the original version of ```scan.py``` I get this:
```
========= 7 failed, 76 passed, 1 skipped, 63 xfailed in 128.90 seconds =========
```
and with the fixed bug, it gives:
```
========= 3 failed, 80 passed, 1 skipped, 63 xfailed in 110.51 seconds =========
```

inducer:
OIC. Thanks for following up.

PR #181, opened 2017-05-19
fix for some cases where the extreme memory sizes cause issues

patrickmmartin:
I observed that depending upon the driver and openCL implementation, the 1GB transfer can fail.  
Rather than attempting to guess (and fail) the correct range, this patch simple handles the failures.


inducer:
Thanks!

PR #212, opened 2017-12-12
Device type 3

kif:
Apparently, the device type is a bit-field, so a device can be both a CPU and a GPU, .... In the debian-CI  test computer the device identifies itself as "Default + CPU"
The idea of this patch is to make the test on the most significant bit only.

### DIFF
--- a/pyopencl/cffi_cl.py
+++ b/pyopencl/cffi_cl.py
@@ -331,7 +331,21 @@ class platform_info(_ConstantsNamespace):  # noqa
 
 
 class device_type(_ConstantsNamespace):  # noqa
-    pass
+    @classmethod
+    def to_string(cls, value, default_format=None):
+        for name in dir(cls):
+            if name in ("DEFAULT", "ALL"):
+                continue
+            if not name.startswith("_"):
+                bitfield = getattr(cls, name)
+                if ((bitfield & value) == bitfield):
+                    return name
+
+        if default_format is None:
+            raise ValueError("a name for value %d was not found in %s"
+                             % (value, cls.__name__))
+        else:
+            return default_format % value
 
 
 class device_info(_ConstantsNamespace):  # noqa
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

aukieranw:
Hi @inducer @kif 

Something in this change seems to not be quite right. Testing with examples/benchmark.py in the master branch gives me the following error;

```
>>> Execution time of test without OpenCL:  0.0529999732971 s
===============================================================
Platform name: NVIDIA CUDA
Platform profile: FULL_PROFILE
Platform vendor: NVIDIA Corporation
Platform version: OpenCL 1.2 CUDA 9.1.75
---------------------------------------------------------------
Device name: GeForce GTX 1070
  File "C:\Python27\Lib\site-packages\Pythonwin\pywin\framework\scriptutils.py", line 323, in RunScript
    debugger.run(codeObject, __main__.__dict__, start_stepping=0)
  File "C:\Python27\Lib\site-packages\Pythonwin\pywin\debugger\__init__.py", line 60, in run
    _GetCurrentDebugger().run(cmd, globals,locals, start_stepping)
  File "C:\Python27\Lib\site-packages\Pythonwin\pywin\debugger\debugger.py", line 654, in run
    exec cmd in globals, locals
  File "C:\Temp\Download\benchmark.py", line 3, in <module>
    from __future__ import print_function
  File "C:\Python27\lib\site-packages\pyopencl\cffi_cl.py", line 346, in to_string
    % (value, cls.__name__))
ValueError: a name for value 4 was not found in device_type
```

Reverting the device_type class back to `pass` resolves the issue (for my use case at least) and the benchmark script then runs without issue. Apologies my knowledge of python is extremely limited otherwise I would have provided a fix.

I have run this test on multiple PC's with quite different hardware and am getting identical results on both.

kif:
On Mon, 18 Dec 2017 04:04:35 +0000 (UTC)
Kieran Wallace <notifications@github.com> wrote:

> Hi @inducer @kif 
> 
> Something in this change seems to not be quite right. Testing with examples/benchmark.py in the master branch gives me the following error;

So it is python 2.7 under windows (32, 64 bits ?). I will investigate on my side. 

The "pass" is just to use the same behavior of the other constants.
Unlike most (all?) other constants, device_type is a bit-field which
needs to be handled specifically.

Cheers

-- 
Jérôme Kieffer
tel +33 476 882 445

aukieranw:
@kif I'm using python 2.7 on win64

Thanks!

kif:
I confirm the issue on windows64 python2.7

kif:
On Python2, "int" and "long" are different, especially on windows.
the test on isinstance(bitfield, int)  --> isinstance(bitfield, (int, long)), but only on python2

kif:
six.integer_types should do the job.

PR #227, opened 2018-03-22
Adds the Multi-threaded Optimization Toolbox

robbert-harms:
Adds MOT to the list of software that builds on the PyOpenCL library.


inducer:
Thanks! From a quick look at your install instructions, you may be interested in [this new conda-based procedure](https://documen.tician.de/pyopencl/misc.html) that gets users up and running with a CPU run time in minimal time and is easily extended to talk to  GPUs, too.

Issue #217, opened 2018-01-05
interop demos both segfault


whbdupree:
My clinfo output: https://pastebin.com/sBhe9ceC

I am noticing that in clinfo, I have way more extensions than show up in pyopencl:

In [15]: p
Out[15]: <pyopencl.Platform 'AMD Accelerated Parallel Processing' at 0x7f1d7bb18510>

In [16]: p.extensions
Out[16]: 'cl_khr_icd cl_amd_event_callback cl_amd_offline_devices '

inducer:
Could you provide tracebacks of the segfaults?

inducer:
Closing for lack of activity and clear reproducible bug. Reopen if this is still an issue.

s-ol:
Seems like this is the same issue I am hitting, running Arch Linux 4.16.8 with a GTX 970.
This happens with my own code as well as the two interop examples from this repository.
Unlike #177 my card does say it supports `cl_khr_gl_sharing` (see clinfo below).
I just saw the half-fix suggested there but won't get to try it until at the earliest tomorrow, the segfault is a bug either way though.

gdb traceback:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007fffecb25c95 in ?? () from /usr/lib/libGLX_nvidia.so.0
(gdb) bt
#0  0x00007fffecb25c95 in ?? () from /usr/lib/libGLX_nvidia.so.0
#1  0x00007fffeb909ccd in ?? () from /usr/lib/libnvidia-glcore.so.396.24
#2  0x00007fffecb4b6a3 in glcuR0d4nX () from /usr/lib/libGLX_nvidia.so.0
#3  0x00007fffe8cb6b49 in ?? () from /usr/lib/libnvidia-opencl.so.1
#4  0x00007fffe8bb1825 in ?? () from /usr/lib/libnvidia-opencl.so.1
#5  0x00007fffe8bb146b in ?? () from /usr/lib/libnvidia-opencl.so.1
#6  0x00007ffff0b32e27 in create_context_from_type () from /usr/lib/python3.6/site-packages/pyopencl/_cffi.abi3.so
#7  0x00007ffff0b1d91c in ?? () from /usr/lib/python3.6/site-packages/pyopencl/_cffi.abi3.so
#8  0x00007ffff7420343 in _PyCFunction_FastCallDict () from /usr/lib/libpython3.6m.so.1.0
[...]

PR #234, opened 2018-05-26
Fix OpenGL interop segfaults on Linux

s-ol:
see issue #217. Some sort of type mismatch/failed cast turned the pointers into signed python integers (hence the negative numbers).
GLX.glXGetCurrentContext() returns a proper python-typed C type that gets casted correctly by the Context.\_\_init\_\_() logic.


inducer:
LGTM. Thanks for working on this!

PR #237, opened 2018-08-05
Array.events handling fixes

rebecca-palmer:
Array.events appears to be intended as "the contents of this Array will be up to date after these", but is currently not used everywhere that that would imply it should be.  Hence, using Array operations in an out-of-order queue (pyopencl.command_queue_properties.OUT_OF_ORDER_EXEC_MODE_ENABLE) can give silently wrong answers.

See the included tests for examples.  Whether they actually fail in current pyopencl is ICD- and possibly hardware-dependent: for me, 3 fail in pocl (CPU), but all pass in beignet (Intel GPU).

This is meant to fix all such places in array/clmath, but not clrandom (fill_rand; rand is already OK) or elementwise/reduce/scan.

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -624,7 +624,7 @@ class Array(object):
             event1 = cl.enqueue_copy(queue or self.queue, self.base_data, ary,
                     device_offset=self.offset,
                     is_blocking=not async_)
-            if not async_: # not already waited for
+            if not async_:  # not already waited for
                 self.add_event(event1)
 
     def get(self, queue=None, ary=None, async_=None, **kwargs):
@@ -1293,7 +1293,8 @@ class Array(object):
         krnl = get_any_kernel(self.context, self.dtype)
         if wait_for is None:
             wait_for = []
-        result, event1 = krnl(self, queue=queue, wait_for=wait_for + self.events, return_event=True)
+        result, event1 = krnl(self, queue=queue,
+               wait_for=wait_for + self.events, return_event=True)
         result.add_event(event1)
         return result
 
@@ -1302,7 +1303,8 @@ class Array(object):
         krnl = get_all_kernel(self.context, self.dtype)
         if wait_for is None:
             wait_for = []
-        result, event1 = krnl(self, queue=queue, wait_for=wait_for + self.events, return_event=True)
+        result, event1 = krnl(self, queue=queue,
+               wait_for=wait_for + self.events, return_event=True)
         result.add_event(event1)
         return result
 
@@ -1686,7 +1688,7 @@ class Array(object):
         if flags is None:
             flags = cl.map_flags.READ | cl.map_flags.WRITE
         if wait_for is None:
-            wait_for=[]
+            wait_for = []
 
         ary, evt = cl.enqueue_map_buffer(
                 queue or self.queue, self.base_data, flags, self.offset,
@@ -2274,7 +2276,6 @@ def multi_put(arrays, dest_indices, dest_shape=None, out=None, queue=None,
     if wait_for is None:
         wait_for = []
     wait_for = wait_for + dest_indices.events
-    
 
     vec_count = len(arrays)
 
@@ -2535,7 +2536,8 @@ def sum(a, dtype=None, queue=None, slice=None):
     """
     from pyopencl.reduction import get_sum_kernel
     krnl = get_sum_kernel(a.context, dtype, a.dtype)
-    result, event1 = krnl(a, queue=queue, slice=slice, wait_for=a.events, return_event=True)
+    result, event1 = krnl(a, queue=queue, slice=slice, wait_for=a.events,
+            return_event=True)
     result.add_event(event1)
     return result
 
@@ -2546,7 +2548,8 @@ def dot(a, b, dtype=None, queue=None, slice=None):
     """
     from pyopencl.reduction import get_dot_kernel
     krnl = get_dot_kernel(a.context, dtype, a.dtype, b.dtype)
-    result, event1 = krnl(a, b, queue=queue, slice=slice, wait_for=a.events + b.events, return_event=True)
+    result, event1 = krnl(a, b, queue=queue, slice=slice,
+            wait_for=a.events + b.events, return_event=True)
     result.add_event(event1)
     return result
 
@@ -2559,7 +2562,8 @@ def vdot(a, b, dtype=None, queue=None, slice=None):
     from pyopencl.reduction import get_dot_kernel
     krnl = get_dot_kernel(a.context, dtype, a.dtype, b.dtype,
             conjugate_first=True)
-    result, event1 = krnl(a, b, queue=queue, slice=slice, wait_for=a.events + b.events, return_event=True)
+    result, event1 = krnl(a, b, queue=queue, slice=slice,
+            wait_for=a.events + b.events, return_event=True)
     result.add_event(event1)
     return result
 
@@ -2572,7 +2576,7 @@ def subset_dot(subset, a, b, dtype=None, queue=None, slice=None):
     krnl = get_subset_dot_kernel(
             a.context, dtype, subset.dtype, a.dtype, b.dtype)
     result, event1 = krnl(subset, a, b, queue=queue, slice=slice,
-        wait_for=subset.events + a.events + b.events, return_event=True)
+            wait_for=subset.events + a.events + b.events, return_event=True)
     result.add_event(event1)
     return result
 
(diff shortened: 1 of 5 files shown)


inducer:
Thanks for your contribution!

Transplanted here for CI: https://gitlab.tiker.net/inducer/pyopencl/merge_requests/43

inducer:
Some linter issues:

https://gitlab.tiker.net/inducer/pyopencl/-/jobs/57300

inducer:
FWIW, I'd be happy to create an account for you on my gitlab instance, to make this less cumbersome.

inducer:
> Array.events appears to be intended as "the contents of  this Array will be up to date after these"

Yes, that's the intention.

> but is currently not used  everywhere that that would imply it should be.

Thank you for helping to fix this feature! It got added to the array interface late, and apparently I missed some (quite a few!) spots.

> This is meant to fix all such places in array/clmath, but not clrandom (fill_rand; rand is already OK) or elementwise/reduce/scan.

Thank you also for doing this survey of other spots that need to be fixed. Could you file issues (don't have to be super wordy-just refer back to this MR) for the stuff that remains to be done?

inducer:
Looks like this causes mysterious crashes when forcing the CL version to 1.1 with POCL: https://gitlab.tiker.net/inducer/pyopencl/-/jobs/57302

rebecca-palmer:
> I'd be happy to create an account for you on my gitlab instance

This may well be a one-off contribution: it's something I found while investigating [these](https://bugs.launchpad.net/ubuntu/+source/beignet/+bug/1354086) beignet [bugs](https://lists.freedesktop.org/archives/beignet/2018-July/009213.html).

> Some linter issues:

One of which looks like an actual bug: wait_for_this shouldn't be unused.

> mysterious crashes

Will investigate later.

> other spots that need to be fixed

I'd assumed algorithms (elementwise/reduce/scan/etc) was *intentionally* "if you use this you're on your own for event management" given its total lack of Array.events use; if this isn't the case, I can have a look at it.

rebecca-palmer:
Pushed fixes for the unused wait_for_this bug and flake8 compliance.

The crash under CL 1.1 is in test_algorithm.py::test_bitonic_sort and is pocl-specific:
\>\>\> sgs, evt = sorter(aCL.copy(), axis=1)
The entire clEnqueueWaitForEvents call is unimplemented ([/build/pocl-i0qN07/pocl-1.1/lib/CL/clEnqueueWaitForEvents.c:39](https://sources.debian.org/src/pocl/1.1-5/lib/CL/clEnqueueWaitForEvents.c/?hl=39#L39))

This is probably an existing issue exposed by my copy() fix, as bitonic sort uses enqueue_marker, which under 1.1 is "if there is anything to wait for, call clEnqueueWaitForEvents".

pyopencl/array.py
```diff
@@ -621,9 +621,11 @@ def set(self, ary, queue=None, async_=None, **kwargs):
                     stacklevel=2)
 
         if self.size:
-            cl.enqueue_copy(queue or self.queue, self.base_data, ary,
+            event1 = cl.enqueue_copy(queue or self.queue, self.base_data, ary,
                     device_offset=self.offset,
                     is_blocking=not async_)
+            if not async_:  # not already waited for
```
inducer: I would actually argue that this event should be added unconditionally, independent of `async_`, for consistency. Compute kernels are also "async" in that sense, and their events end up in the list.

inducer: While we're at it--could you add a sentence to the docstring documenting this behavior?

inducer:
Looks great, CI passes. Thanks for the contribution! Good to go in once the two minor semantics/doc concerns I mentioned are addressed.

pyopencl/array.py
rebecca-palmer: That's actually a bug - as noted in the comment, it's meant to be an "only add events we haven't already waited for" optimization, but that's "if async_", i.e. it's the wrong way round.

Which suggests this optimization isn't worth the risk of creating bugs, so the event is now unconditionally added.

inducer:
Thanks for your contribution!

rebecca-palmer:
I have now extended this to cover clrandom and most algorithms.

Still excluded:
- ElementwiseKernel and GenericScanKernel/GenericDebugScanKernel wait for their arguments' events, but don't add their event to their output, because they don't know which argument(s) are the output.  (The more specific scan-based operations do know this, so do add their event.) Options: 
  - Leave them as they are and document this.
  - Add their event to all arguments and accept the potential performance penalty of unnecessary waits.
  - Extend the argument list parsing to look for 'const'=input-only labels.
- ListOfListsBuilder can't wait because it takes bare Buffer arguments not Array arguments.  It currently also doesn't add its event to new Arrays it creates.

Also, before we start promoting the use of out-of-order-queues, I suspect there may be another problem, with MemoryPool: it looks possible (though I haven't yet tried to create an example) for an array to be garbage collected, its memory returned to the pool and re-used as a different array while it still has events pending, causing the new array to get overwritten by the old kernel.

inducer:
Thanks for raising these concerns! My preferred solution would be:

* **By default:** Add their event to all arguments and accept the potential performance penalty of unnecessary waits.
* **Optionally:** Extend the argument list parsing to look for 'const'=input-only labels.
*   > ListOfListsBuilder can't wait because it takes bare Buffer arguments not Array arguments.

    Hmm, that's dumb. I don't recall there being a good reason other than laziness. I imagine we should deprecate bare buffers and move towards accepting arrays. (And state that, in the case of bare buffers, the user alone is responisble for making sure `wait_for` is correct.)

*   As for the mempool concern, I agree that that's a possibility. IMO, we should simply wait for all events in the array destructor.

rebecca-palmer:
Elementwise / scan add event to all arguments: done.  (GitHub evidently doesn't show new commits to an already-merged request.)

Documenting that using bare Buffers bypasses event management: done.

MemoryPool: waiting in Array.\_\_del\_\_ wouldn't be enough, as only Arrays that are being (possibly)
**written to** get Events, and it's also a problem if an operation **reads from** memory that has been reused as a different object.

Hence, this instead makes kernel Events into NannyEvents that guard all their CLobject arguments (in and out).  To avoid tying up potentially large amounts of memory for longer than necessary, it also makes (all) NannyEvents release on completion if possible, instead of waiting until they are destroyed or explicitly wait()ed.  Warning: only lightly tested.

For similar reasons, copy NannyEvents should probably guard *both* buffers; I haven't yet tried to implement this.

inducer:
Very cool! Thanks again for your work on this.

> (GitHub evidently doesn't show new commits to an already-merged request.)

Could you just push that to a fresh branch with a fresh pull request? I think that will be a better fit to Github's workflow.

On a semi-related note, #240 will supersede master in the relatively short term--so to save trouble on merges later, I think it would be best if you aimed your fixes at that branch (`pybind11`). (Bonus: That branch is also easier to hack--at least I think so.)

> MemoryPool: waiting in Array.__del__ wouldn't be enough, as only Arrays that are being (possibly)
written to get Events, and it's also a problem if an operation reads from memory that has been reused as a different object.

Ugh, you're right. That's tricky. Maybe we need an `Array.reading_events`, which would get waited on by any in-place modification. (There aren't that many of those.) The main challenge would be keeping the size of that list under control.

> Hence, this instead makes kernel Events into NannyEvents that guard all their CLobject arguments (in and out). To avoid tying up potentially large amounts of memory for longer than necessary, it also makes (all) NannyEvents release on completion if possible, instead of waiting until they are destroyed or explicitly wait()ed. Warning: only lightly tested.

I don't see how that prevents re-adding a buffer to a pool before every kernel is done reading it. Could you explain?

If we decide we want this, I think the from-buffer-reading copies would also need this, for consistency.

rebecca-palmer:
> Ugh, you're right. That's tricky. Maybe we need an Array.reading_events, which would get waited on by any in-place modification. (There aren't that many of those.) The main challenge would be keeping the size of that list under control.

You're right - it's not just MemoryPool return-to-pool, it's ordering of operations like b=a*a; a+=5.

>  >   Hence, this instead makes kernel Events into NannyEvents that guard all their CLobject arguments (in and out). [...]

> I don't see how that prevents re-adding a buffer to a pool before every kernel is done reading it. Could you explain?

The idea is that having a Python reference to the argument (which will be a PooledBuffer object if it's from a pool) prevents PooledBuffer.\_\_del\_\_ being called, but I still haven't tried to create an actual test case for this issue.

> If we decide we want this, I think the from-buffer-reading copies would also need this, for consistency.

Pushed that because I'd already written it, but given the above, I suggest *not* accepting it for the time being.

Two unrelated maybe-bugs noticed while writing this:
- enqueue_migrate_mem_object_ext (cffi_cl.py:2037 in this branch) calls the same-name C function, which does not appear to exist.
- enqueue_svm_**migrate**mem (cffi_cl.py:1452) calls C enqueue_svm_mem**fill**, which makes no sense.

inducer:
> operations like b=a*a; a+=5.

Yeah. Although `b=a*a` is not in-place and hence fine.

> The idea is that having a Python reference to the argument (which will be a PooledBuffer object if it's from a pool) prevents PooledBuffer.__del__ being called

OIC

> enqueue_migrate_mem_object_ext (cffi_cl.py:2037 in this branch) calls the same-name C function, which does not appear to exist.
> enqueue_svm_migratemem (cffi_cl.py:1452) calls C enqueue_svm_memfill, which makes no sense.

Thanks for pointing these out. I've merged the pybind11 branch in the meantime, and I think these two are fixed/no longer relevant there.

PR #238, opened 2018-08-05
Intel GPU (beignet-opencl-icd) compatibility fixes

rebecca-palmer:
Fix int64 scan crash, and skip tests that exceed hardware capabilities.  (beignet is a 2.0 platform but on older hardware only a 1.2 device.)

Some of these changes take effect only after clearing the caches (rm -rf  ~/.cache/pyopencl ~/.cache/pytools/\*pyopencl\* ).

Ranlux int64 still fails its tests, but that's deprecated anyway.

Something this does _not_ fix, partly because I'm not sure at what level it would be appropriate to do so: zero-size host-to-device copies (bCL[8:8]=a[8:8]) fail with pyopencl.cffi_cl.LogicError: clEnqueueWriteBuffer failed: INVALID_VALUE

### DIFF
--- a/pyopencl/scan.py
+++ b/pyopencl/scan.py
@@ -1227,8 +1227,12 @@ class GenericScanKernel(_GenericScanKernelBase):
             max_scan_wg_size = min(dev.max_work_group_size for dev in self.devices)
             wg_size_multiples = 64
 
+        # Intel beignet asserts or gives wrong results with packed structs
+        # https://bugs.freedesktop.org/show_bug.cgi?id=98717
+        # TODO: is this all Intel ICDs or only beignet?
         use_bank_conflict_avoidance = (
-                self.dtype.itemsize > 4 and self.dtype.itemsize % 8 == 0 and is_gpu)
+                self.dtype.itemsize > 4 and self.dtype.itemsize % 8 == 0
+                and is_gpu and "Intel" not in self.devices[0].platform.name)
 
         # k_group_size should be a power of two because of in-kernel
         # division by that number.
--- a/pyopencl/scan.py
+++ b/pyopencl/scan.py
@@ -1227,12 +1227,17 @@ class GenericScanKernel(_GenericScanKernelBase):
             max_scan_wg_size = min(dev.max_work_group_size for dev in self.devices)
             wg_size_multiples = 64
 
-        # Intel beignet asserts or gives wrong results with packed structs
+        # Intel beignet fails "Out of shared local memory" in test_scan int64
+        # and asserts in test_sort with this enabled:
+        # https://github.com/inducer/pyopencl/pull/238
+        # A beignet bug report (outside of pyopencl) suggests packed structs
+        # (which this is) can even give wrong results:
         # https://bugs.freedesktop.org/show_bug.cgi?id=98717
-        # TODO: is this all Intel ICDs or only beignet?
+        # TODO: does this also affect Intel Compute Runtime?
         use_bank_conflict_avoidance = (
                 self.dtype.itemsize > 4 and self.dtype.itemsize % 8 == 0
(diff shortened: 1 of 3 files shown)


inducer:
Thanks for your contribution!

Transplanted here for CI: https://gitlab.tiker.net/inducer/pyopencl/merge_requests/42

pyopencl/scan.py
```diff
@@ -1227,8 +1227,12 @@ def _finish_setup_impl(self):
             max_scan_wg_size = min(dev.max_work_group_size for dev in self.devices)
             wg_size_multiples = 64
 
+        # Intel beignet asserts or gives wrong results with packed structs
+        # https://bugs.freedesktop.org/show_bug.cgi?id=98717
+        # TODO: is this all Intel ICDs or only beignet?
```
inducer: I think this is a tad too broad--their CPU ICD reports

```
===========================================================================
<pyopencl.Platform 'Intel(R) OpenCL' at 0x1e02e80>
===========================================================================
EXTENSIONS: cl_khr_icd cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_byte_addressable_store cl_khr_depth_images cl_khr_3d_image_writes cl_intel_exec_by_local_thread cl_khr_spir cl_khr_fp64 cl_khr_image2d_from_buffer 
NAME: Intel(R) OpenCL
PROFILE: FULL_PROFILE
VENDOR: Intel(R) Corporation
VERSION: OpenCL 2.0 LINUX
-------------------------------
```

pyopencl/scan.py
```diff
@@ -1227,8 +1227,12 @@ def _finish_setup_impl(self):
...
         use_bank_conflict_avoidance = (
-                self.dtype.itemsize > 4 and self.dtype.itemsize % 8 == 0 and is_gpu)
+                self.dtype.itemsize > 4 and self.dtype.itemsize % 8 == 0
+                and is_gpu and "Intel" not in self.devices[0].platform.name)
```
inducer: Same as above, too broad.

inducer (commented):
LGTM aside from these two issues.

rebecca-palmer:
The "this" of "TODO: is this all Intel ICDs or only beignet?" is the packed struct bug that requires use_bank_conflict_avoidance to be off: it's intentional that the current check disables it for all Intel ICDs, as without testing that's the safe option.

(There's also [Intel Compute Runtime](https://01.org/compute-runtime), which requires better hardware than I have.)

inducer:
OIC. The Intel CPU ICD is not affected by the struct arg count bug as far as I know.

There's also [this function](https://github.com/inducer/pyopencl/blob/7c2f3ba43337b98e57db06d2e7b82de3e07266ae/pyopencl/characterize/__init__.py#L375) whose job it is to answer the question whether an ICD is affected.

rebecca-palmer:
beignet isn't affected by [that arg count bug](https://github.com/pocl/pocl/issues/197) either: it counts 1 2-entry struct (packed or not) as 1 argument.  Hence, this must be a different issue.

Does the existing test suite (_without_ this pull request), in particular test_scan's int64 cases and test_sort (fail or crash on beignet), pass on the other Intel ICDs?  Your CI script says yes on Intel CPU, but I'm more worried about Intel Compute Runtime (Broadwell-and-later GPU).

If they pass, the test can become `and "beignet" not in self.devices[0].platform.version.lower())` (yes, it's in the version not the name).

inducer:
I don't have a way of testing right now. But I think it's fair to only cover those ICDs with workarounds for which we have a means of testing.

inducer:
Thanks for seeing this through!

PR #241, opened 2018-09-11
Fix bank conflict avoidance in scan

adityapb:
Shouldn't `ldata` be padded in `WG_SIZE` regardless of whether padding is required for `wrapped_scan_type`.
This gives me about 2x speedup using `np.int32` dtype

### DIFF
--- a/pyopencl/scan.py
+++ b/pyopencl/scan.py
@@ -157,7 +157,6 @@ void ${kernel_name}(
             scan_type value;
             int dummy;
         };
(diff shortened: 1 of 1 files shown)


inducer:
Transplanted here for CI: https://gitlab.tiker.net/inducer/pyopencl/merge_requests/46

inducer:
You're right. I must have gotten confused between the two different kinds of bank conflict avoidance. Thanks for your contribution!

PR #249, opened 2018-10-11
removed exception "CommandQueue failed: INVALID_VALUE"

OmegaDoom:
Hi.
```
properties = cl.command_queue_properties.OUT_OF_ORDER_EXEC_MODE_ENABLE\
            | cl.command_queue_properties.ON_DEVICE
queue_properties = [cl.queue_properties.PROPERTIES, properties,
                    cl.queue_properties.SIZE, 1000000]
device_queue = cl.CommandQueue(context, device, queue_properties)
```

This code throws exception - "pyopencl._cl.LogicError: CommandQueue failed: INVALID_VALUE - queue properties given as an iterable, which is only allowed when PyOpenCL was built against an OpenCL 2+ header"

I removed the exception and code works fine. Does this exception have value or added by mistake?

### DIFF
--- a/src/wrap_cl.hpp
+++ b/src/wrap_cl.hpp
@@ -1356,11 +1356,6 @@ namespace pyopencl
         else
         {
 #if PYOPENCL_CL_VERSION >= 0x2000
-            throw error("CommandQueue", CL_INVALID_VALUE,
-                "queue properties given as an iterable, "
-                "which is only allowed when PyOpenCL was built "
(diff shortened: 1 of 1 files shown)


inducer:
Good catch, but not the right fix. Closing in favor of https://gitlab.tiker.net/inducer/pyopencl/merge_requests/58.

inducer:
Thanks for reporting this!

Issue #196, opened 2017-07-27
vbo.buffers[0] is not a valid accessor in PyOpenGL 3.1


fizyxnrd:
This is an issue in the gl_particle_animation example (lines 146, 147) which refer to
    int(gl_position.buffers[0])
    int(gl_color.buffers[0])

The code does work with
    int(gl_position.buffer)
    int(gl_color.buffer)

inducer:
Thanks for the report. Could you try the fixed version and confirm that it works?

mamat-rahmat:
@fizyxnrd @inducer Seems like the fixed version not work.

```
Traceback (most recent call last):
  File "03_particle.py", line 146, in <module>
    cl_gl_position = cl.GLBuffer(context, mf.READ_WRITE, int(gl_position.buffer))
  File "C:\Program Files\Python37\lib\site-packages\OpenGL\arrays\vbo.py", line 277, in __getattr__
    return getattr( self.data, key )
AttributeError: 'numpy.ndarray' object has no attribute 'buffer'
```

I changed to previous version then its worked

inducer:
@mamat-rahmat What version of PyOpenGL were you using?

mamat-rahmat:
PyOpenGL==3.1.3b2

inducer:
So, if I understand you correctly, you're essentially suggesting we revert 7866408? Can you point to a change in PyOpenGL (between 3.1 and 3.1.3) that this relates to? @fizyxnrd, any opinion on this? (I don't currently have GL-interop-capable hardware to test with.)

mamat-rahmat:
Recently I found that I didn't install `PyOpenGL_accelerate` (i use 3.1.3b2)

So, the situation is, if `PyOpenGL_accelerate` installed, `int(gl_position.buffer)` works
Else, it works with `int(gl_color.buffers[0])`

How about using both, i mean, like
```
try:
    # use buffer
except AttributeError:
    # use buffers[0]
```

I'm now investigating the cause

mamat-rahmat:
> Basically the code is trying to access the "buffers" member of the VBO object. The OpenGL_accelerate version of the VBO (the cython-implemented one) does *not* have a .buffers (list), only a .buffer (int). Probably the attributes should be named _buffer and/or _buffers, as they aren't normally used by external code, the normal way to get the ID of the VBO is int(<vbo>).

Source : https://sourceforge.net/p/pyopengl/mailman/message/32883960/

So, the safest way is just use
```
[...]

PR #259, opened 2018-11-20
Make particle example work without PyOpenGL_accelerate

mamat-rahmat:
vbo.buffer nor vbo.buffers aren't normally used by external code
The normal way to get the ID of the VBO is int(vbo)
See https://github.com/inducer/pyopencl/issues/196


inducer:
Thanks!

Issue #250, opened 2018-10-22
Mark more characters as safe when passing build options


SemMulder:
Right now the characters which are marked as safe [here](https://github.com/inducer/pyopencl/blob/4c0521913881ac6e284b38c9aab79708dfb522db/pyopencl/__init__.py#L392), are not exhaustive. Since we aren't actually passing the build options through a shell, but are invoking functions like `clBuildProgram`, the characters `-_{}()` are also safe.

I ran into this when compilation failed when trying to pass an array literal using `-D ARRAY_LITERAL={0,1}`. This just works if we mark the mentioned characters as safe.

I have a fix living [here](https://github.com/SemMulder/pyopencl), for which I can't open a PR on https://gitlab.tiker.net/inducer/pyopencl since I don't have an account.

Thoughts?

inducer:
Thanks for the report and the suggested fix.

What'll happen if "unsafe" characters are found is that the affected argument gets quoted. Are you saying that the quoting is breaking not working in your use case.

SemMulder:
Precisely! So the quoting results in the following:
```
<built-in>:9:9: error: macro name must be an identifier
In file included from stringInput.cl:2:
main.cl:193:57: error: use of undeclared identifier 'NEIGHBOURHOOD_ARRAY'
main.cl:348:57: error: use of undeclared identifier 'NEIGHBOURHOOD_ARRAY'

(options: ... -D 'NEIGHBOURHOOD_ARRAY={{-1,0},{0,-1},{0,1},{1,0}}' ...)
```

Note the quotes in the output. This happens with every OpenCL runtime I tested, so I don't think it's a bug in the runtime.

I checked the usages of [`_process_build_options`](https://github.com/inducer/pyopencl/blob/master/pyopencl/__init__.py#L417) (which is the only usage of [`_shlex_qoute`](https://github.com/inducer/pyopencl/blob/master/pyopencl/__init__.py#L395). They are
- https://github.com/inducer/pyopencl/blob/master/pyopencl/__init__.py#L475,
- https://github.com/inducer/pyopencl/blob/master/pyopencl/__init__.py#L583,
- https://github.com/inducer/pyopencl/blob/master/pyopencl/__init__.py#L559.

None of them need _shell_ escaping, as far as I can tell. Since they all call `clBuildProgram` and `clLinkProgram` directly (i.e. not via a shell).

[...]

PR #261, opened 2018-11-27
Delegate quoting of build options to the user

SemMulder:
This PR addresses #250. We simply don't try to quote anything from the user, and we quote (and unquote) the default include paths where necessary.


inducer:
Replicated for CI at https://gitlab.tiker.net/inducer/pyopencl/merge_requests/65

PR #284, opened 2019-03-30
Add cfloat_t and cdouble_t to cltypes

chatcannon:
If I understand correctly, it is not possible to create vectors of cfloat_t or cdouble_t at the moment, so I have not added any corresponding vector types.

### DIFF
--- a/pyopencl/cltypes.py
+++ b/pyopencl/cltypes.py
@@ -42,6 +42,8 @@ ulong = np.uint64
 half = np.float16
 float = np.float32
 double = np.float64
+cfloat_t = np.complex64
+cdouble_t = np.complex128
 
 
 # {{{ vector types


inducer:
Are these standard OpenCL data types? The [OpenCL C 2.0 spec](https://www.khronos.org/registry/OpenCL/specs/2.2/pdf/OpenCL_C.pdf#page=15) specifically lists complex types as reserved.

chatcannon:
No, they're not in the standard, but since they are in pyopencl's pyopencl-complex.h I thought it would be helpful to add them here.

inducer:
The types there are only supposed to reflect ones in the standard (at least as documented). So, thanks for your contribution, but it's not currently consistent with the design. Sorry!

PR #298, opened 2019-09-19
docs(README.rst): embrace https links

yxliang01:
Changed links that were in http protocol and confirmed supporting https to https.


inducer:
Thanks! I'll wait for CI to pass for the green checkmark's sake, but this should be good to go.

PR #306, opened 2019-11-05
Rebuild from source if cached version causes BUILD_PROGRAM_FAILURE

pkienzle:
Issue #276 can be worked around by rebuilding a program from source if it can't be built from cache.

This doesn't need to be OS or driver specific: if any driver is failing to rebuild from cache it doesn't hurt to rebuild from source.

We may want to remove the bad cache entry when this happens so we are not forced to rebuild from source even after the driver has been fixed, but that is a much bigger change.


inducer:
This looks like a pretty desperate measure; you're paying for a from-source build on every kernel load. Have you checked out pocl on mac (easy to install via conda, see install instructions)?

At any rate, I guess it does beat having the package not work, and the logic looks OK to me. Thanks for your contribution!

PR #316, opened 2019-12-04
Fix slow import of pyopencl

cgohlke:
On my Windows workstation, `import pyopencl` takes 3.2 s, more than 3 s of which are spent in `import pkg_resources`.

Importing `pkg_resources` is known to be excruciatingly slow. See https://github.com/pypa/setuptools/issues/510

This PR just reverses the order of the methods used to find the cl include directory. I'm not sure if `pkg_resources` is really necessary here at all (?)


inducer:
Thanks for the contribution! I'd be happy to take this patch if you fix the [linter failure](https://dev.azure.com/ak-spam/inducer/_build/results?buildId=444&view=logs).

PR #338, opened 2020-03-30
added support for arrays as struct fields

sniarchos:
I tried to add as little code as possible. I think that now arrays (of as many dimensions as the user wants) as struct fields are supported.

### DIFF
--- a/pyopencl/tools.py
+++ b/pyopencl/tools.py
@@ -463,6 +463,13 @@ class _CDeclList:
         if dtype in pyopencl.cltypes.vec_type_to_scalar_and_count:
             return
 
+        if hasattr(dtype, "subdtype") and dtype.subdtype is not None:
+            if dtype.subdtype[0] in [np.float64 or np.complex128]:
+                self.saw_double = True
+            if dtype.subdtype[0].kind == "c":
+                self.saw_complex = True
+            return
+
         for name, field_data in sorted(six.iteritems(dtype.fields)):
             field_dtype, offset = field_data[:2]
             self.add_dtype(field_dtype)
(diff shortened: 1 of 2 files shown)


inducer:
Could you add a test?

sniarchos:
on it

sniarchos:
Would you like something more elaborate than that? I did not add something like running a kernel with structs because the problem was in `match_dtype_to_c_struct` and `add_dtype`; after these 2 are successful, everything proceeds as normal.

sniarchos:
Nevermind; I added a full test (i.e. with a kernel that uses a struct with array fields)

pyopencl/tools.py
```diff
@@ -463,6 +463,13 @@ def add_dtype(self, dtype):
         if dtype in pyopencl.cltypes.vec_type_to_scalar_and_count:
             return
 
+        if hasattr(dtype, "subdtype") and dtype.subdtype is not None:
+            if dtype.subdtype[0] in [np.float64 or np.complex128]:
```
inducer: This should probably recurse instead of inlining a partial check.

inducer: (Alternatively, check that this is a primitive `dtype` and raise `NotImplementedError` otherwise.)

pyopencl/tools.py
```diff
@@ -547,7 +554,20 @@ def match_dtype_to_c_struct(device, name, dtype, context=None):
...
+            array_dims = field_dtype.subdtype[1]
+            dims_str = ""
+            try:
+                for dim in array_dims:
+                    dims_str += "[%d]" % dim
+            except TypeError:
+                dims_str = "[%d]" % array_dims
+            c_fields.append("  %s %s%s;" % (
+                dtype_to_ctype(array_dtype), field_name, dims_str)
+            )
```
inducer, line 567: If the `subdtype` is another array, will this compose the declarator correctly? If not, check for unhandled cases and raise `NotImplementedError` in those.

inducer (changes requested):
Thanks for your contribution! Some small requests below.

pyopencl/tools.py
sniarchos: done it recursively

pyopencl/tools.py
sniarchos: raised a `NotImplementedError` (sidenote: nested array dtypes are meaningless in my opinion, if the user wants to define a multidimensional array they can do it in one level, i.e. the way I have done it in the test I added)

sniarchos:
something went wrong with the tests that **I think** is not my fault:
```
In file included from src/wrap_constants.cpp:30:0:
src/wrap_cl.hpp:57:10: fatal error: CL/cl.h: No such file or directory
 #include <CL/cl.h>
          ^~~~~~~~~
compilation terminated.
error: command 'gcc' failed with exit status 1
```

inducer (approved):
Thanks for being responsive to my suggestions, and for your contribution overall. LGTM now, I'll investigate what happened with the CI. For now, I've simply rerun the jobs on the assumption that something might be wrong on the conda end.

inducer:
See #340.

PR #340, opened 2020-04-03
Struct array

inducer:
Duplicate of #338, as I was unable to push to that.


sniarchos:
Hello again! Is there any chance that `pyopencl` gets a new release on PyPI with this new added feature? I am asking because I am using it in a project of mine and I am currently depending on a local fork, which I would love to get rid of

inducer:
Done.

PR #375, opened 2020-09-09
Fix, test usage of read-only SVM

inducer:


inducer:
See also #379 for important follow-up work.

PR #382, opened 2020-09-17
Fix platform detection for pocl

matthiasdiener:
In pocl, "dev.platform.vendor" is "The pocl project".

### DIFF
--- a/pyopencl/characterize/__init__.py
+++ b/pyopencl/characterize/__init__.py
@@ -282,11 +282,11 @@ def get_simd_group_size(dev, type_size):
 
     lc_plat_vendor = dev.platform.vendor.lower()
     lc_dev_vendor = dev.vendor.lower()
-    if "nvidia" in lc_vendor or "nvidia" in lc_real_vendor:
+    if "nvidia" in lc_plat_vendor or "nvidia" in lc_dev_vendor:
         return 32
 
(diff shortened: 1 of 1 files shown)


pyopencl/characterize/__init__.py
```diff
@@ -281,12 +281,14 @@ def get_simd_group_size(dev, type_size):
         pass
 
     lc_vendor = dev.platform.vendor.lower()
-    if "nvidia" in lc_vendor:
+    lc_real_vendor = dev.vendor.lower()
```
inducer:
```suggestion
    lc_dev_vendor = dev.vendor.lower()
```

pyopencl/characterize/__init__.py
```diff
@@ -281,12 +281,14 @@ def get_simd_group_size(dev, type_size):
         pass
 
     lc_vendor = dev.platform.vendor.lower()
```
inducer:
```suggestion
    lc_plat_vendor = dev.platform.vendor.lower()
```

inducer (commented):
Thanks, good catch. I do object to the variable names though. :)

matthiasdiener:
> Thanks, good catch. I do object to the variable names though. :)

Fixed, thanks.

inducer:
Thanks!

PR #388, opened 2020-10-14
pipe: make it compile with OpenCL < 2

matthiasdiener:
`cl_pipe_info` doesn't exist in OpenCL < 2, leading to a compilation error.

Related to #376.


inducer:
Thx!

PR #396, opened 2020-11-01
Don't use enqueue_fill_buffer for large arrays on NVIDIA

### LINKED ISSUE #395: enqueue_fill_buffer issues on NVIDIA for nbytes >= 2**31
I'm not sure whether this is a problem in the wrapper or just in NVIDIA's driver, but calls to `enqueue_fill_buffer` (e.g., by [`pyopencl.array.zeros`](https://github.com/inducer/pyopencl/blob/4e2e4f3150c331680e6d9e36c59290411e4a0c40/pyopencl/array.py#L1287)) either hang or result in errors when `nbytes >= 2**31`. On Pascal the call just hangs and on Volta the call returns but throws an error at the OpenCL API call (e.g., clBuildProgram failure with <unknown error -9999>), so far as I can tell. (I've reproduced this on numerous machines and driver versions.)

According to the [spec](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#_footnotedef_4) the size parameter is supposed to be a 64 bit integer (based on `device.address_bits`), but alas.

```python
import pyopencl as cl
import pyopencl.array as cla
queue = cl.CommandQueue(ctx)

print(ctx.devices[0].address_bits) # prints 64 on NVIDIA, so size_t should be 64 bit

x = cla.empty(queue, shape=(2**28,), dtype='float64')
x[...] = 0.
print("I make it here")

y = cla.zeros(queue, shape=(2**28,), dtype='float64')
print("I don't make it here")
```

Regardless of the source of the error, a simple workaround is to just use `Array.fill` instead whenever the platform is NVIDIA and `nbytes >= 2**31`. Is that an acceptable solution?

zachjweiner:
Closes #395.

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -1285,7 +1285,8 @@ class Array:
         )
         on_nvidia = queue.device.vendor.startswith("NVIDIA")
 
-        # circumvent bug with large buffers on NVIDIA (gh-395)
+        # circumvent bug with large buffers on NVIDIA
+        # https://github.com/inducer/pyopencl/issues/395
         if cl_version_gtr_1_2 and not (on_nvidia and self.nbytes >= 2**31):
             self.add_event(
                     cl.enqueue_fill_buffer(queue, self.base_data, np.int8(0),
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -1279,10 +1279,14 @@ class Array:
         if not self.size:
             return
 
-        if (
-                queue._get_cl_version() >= (1, 2)
-                and cl.get_cl_header_version() >= (1, 2)):
+        cl_version_gtr_1_2 = (
+            queue._get_cl_version() >= (1, 2)
+            and cl.get_cl_header_version() >= (1, 2)
+        )
+        on_nvidia = queue.device.vendor.startswith("NVIDIA")
 
+        # circumvent bug with large buffers on NVIDIA (gh-395)
+        if cl_version_gtr_1_2 and not (on_nvidia and self.nbytes >= 2**31):
             self.add_event(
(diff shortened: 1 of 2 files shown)


pyopencl/array.py
```diff
@@ -1279,10 +1279,14 @@ def _zero_fill(self, queue=None, wait_for=None):
...
-                queue._get_cl_version() >= (1, 2)
-                and cl.get_cl_header_version() >= (1, 2)):
-
+        cl_version_gtr_1_2 = (
+            queue._get_cl_version() >= (1, 2)
+            and cl.get_cl_header_version() >= (1, 2)
+        )
+        on_nvidia = queue.device.vendor.startswith("NVIDIA")
+
+        # circumvent bug with large buffers on NVIDIA (gh-395)
```
```suggestion
        # circumvent bug with large buffers on NVIDIA
        # https://github.com/inducer/pyopencl/issues/395
```
(Clearer and directly clickable)

inducer (commented):
Thanks! https://gitlab.tiker.net/inducer/pyopencl/-/merge_requests/129 checks on the one Nv device we currently have.

zachjweiner:
CI looks good. Er, should I add a test? (Can we count on 2GB being available?)

inducer:
Yes to both

zachjweiner:
Added a test that should pick up the behavior I've observed. It passes on my local P100 and on GitLab.

zachjweiner:
Had to add a check that the test allocation wouldn't exceed `device.max_mem_alloc_size`.

But the problem CI job (pytest linux) ran into a conda http timeout - would you mind telling github to rerun the pipeline?

inducer (approved):
Thanks!

PR #400, opened 2020-11-16
Add note to install docs about mixing conda forge and pypi installations

### LINKED ISSUE #399: ocl-icd installed via conda replaces previous loader used by pip-installed PyOpenCL
Installing PyOpenCL via pip first links to the `libOpenCL.so.1` already present on my system (installed with CUDA, in my case). After a `conda install ocl-icd`, `ldd` instead says PyOpenCL's shared library points to the `libOpenCL.so.1` installed by `ocl-icd`. This results in, e.g., system icd files no longer appearing as available platforms. 

I'm not sure whether this is expected (or whether mixing pip-installed PyOpenCL with conda-installed loaders is ill-advised). If so, it might be useful to note in the installation docs (happy to contribute).

(FWIW, I discovered this by accident - for reasons I cannot determine as I can no longer reproduce, `ocl-icd-system` was leading to segfaults at `get_platforms()`, and in testing I explored installing PyOpenCL with pip rather than conda.)

zachjweiner:
Per title. Closes #399.

While I'm here, I propose two more additions (for installations in a conda environment):

1. `/WHERE/YOU/INSTALLED/MINICONDA/envs/ENVIRONMENTNAME/` can be obtained by `$CONDA_PREFIX` (for the active environment)
2. As an alternative to installing `ocl-icd-system`, system ICD files can be manually copied from `/etc/OpenCL/vendors` into `$CONDA_PREFIX/etc/OpenCL/vendors/`.

Thoughts?

### DIFF
--- a/doc/misc.rst
+++ b/doc/misc.rst
@@ -94,6 +94,10 @@ qualified path names of the shared library providing the OpenCL driver.
 
     Note that you should replace ``ENVIRONMENTNAME`` with the name of your environment,
     shown between parentheses on your command line prompt.
+    This path (for the currently-active conda environment) can be obtained from the
+    environment variable ``CONDA_PREFIX``, i.e.,
+
+    :file:`$CONDA_PREFIX/etc/OpenCL/vendors`.
 
 On Linux, if you have other OpenCL drivers installed (such as for your GPU), those will be
 in :file:`/etc/OpenCL/vendors`. You can make them work with PyOpenCL from Conda Forge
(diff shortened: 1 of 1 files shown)


inducer:
I'm OK with both. (As long as 2 is not recommended, but mentioned as an alternative.)

zachjweiner:
Done, lmk if the language isn't to your liking.

doc/misc.rst
```diff
@@ -94,14 +94,20 @@ qualified path names of the shared library providing the OpenCL driver.
 
     Note that you should replace ``ENVIRONMENTNAME`` with the name of your environment,
     shown between parentheses on your command line prompt.
+    This path (for the currently-active conda environment) can be obtained from the
+    environment variable ``CONDA_PREFIX``, i.e.,
+
+    :file:`$CONDA_PREFIX/etc/OpenCL/vendors`.
```
inducer:
```suggestion
    :file:`$CONDA_PREFIX/etc/OpenCL/vendors` (once the Conda environment is activated).
```

inducer (approved):

inducer:
Thanks!

PR #405, opened 2020-11-21
Add arm64 support for Apple Silicon

geohot:
Simple change. Let's see if CI passes, or I can gate it based on the local CPU arch.


inducer:
Thanks! Some more CI: https://gitlab.tiker.net/inducer/pyopencl/-/merge_requests/130

inducer:
Looks good. Thanks!

PR #414, opened 2020-12-31
Fix rng evt mgmt

### LINKED ISSUE #413: Unsafe use of fill_uniform in test_array.py
Forgive me if I'm misunderstanding something, but I think [this](https://github.com/inducer/pyopencl/blob/master/test/test_array.py#L745) use of `fill_uniform` immediately preceded by a `.get()` call on the array can result in a data race, since `fill_uniform` invokes the `__call__` operator of `kernel` and returns the event without waiting for it.

inducer:
Closes #413.


aarongreig:
that fixed the issue I was having, thanks for the speedy turnaround!

PR #416, opened 2021-01-15
Speed up enqueue

inducer:
- [x] Gitlab sister MR: https://gitlab.tiker.net/inducer/pyopencl/-/merge_requests/134

cc @matthiasdiener

PR #417, opened 2021-01-19
ARRAY_KERNEL_EXEC_HOOK: estimate bytes/flops

matthiasdiener:


inducer:
I don't think this belongs here. One important reason is that we don't really have a way of looking inside the kernel here, and therefore the `nops` generated is, at best, an estimate, and at worst, dead wrong.

Once #416 is in, the arrays themselves will appear as arguments to the kernel, which should make it possible to do this within the hook itself, which makes it more palatable to me.

Does that sound reasonable?

inducer (changes requested):
I don't think this belongs here. One important reason is that we don't really have a way of looking inside the kernel here, and therefore the `nops` generated is, at best, an estimate, and at worst, dead wrong.

Once #416 is in, the arrays themselves will appear as arguments to the kernel, which should make it possible to do this within the hook itself, which makes it more palatable to me.

Does that sound reasonable?

matthiasdiener:
> I don't think this belongs here. One important reason is that we don't really have a way of looking inside the kernel here, and therefore the `nops` generated is, at best, an estimate, and at worst, dead wrong.
> 
> Once #416 is in, the arrays themselves will appear as arguments to the kernel, which should make it possible to do this within the hook itself, which makes it more palatable to me.
> 
> Does that sound reasonable?

Sure, that sounds fine!

matthiasdiener:
Abandoned in favor of implementing this directly in mirgecom: github.com/illinois-ceesd/mirgecom/pull/193

PR #424, opened 2021-01-29
add DOMAIN_ID to nvidia device info extension

matthiasdiener:
Needs https://github.com/pocl/pocl/pull/910


doc/make_constants.py
```diff
@@ -112,6 +112,7 @@ def get_extra_lines(tup):
             "ATTRIBUTE_ASYNC_ENGINE_COUNT_NV": nv_devattr,
             "PCI_BUS_ID_NV": nv_devattr,
             "PCI_BUS_SLOT_NV": nv_devattr,
+            "PCI_BUS_DOMAIN_NV": nv_devattr,
```
matthiasdiener: It's a bit strange that "ID" is missing in these constants.

matthiasdiener: Wait, why is this even called `PCI_BUS_SLOT_NV`?

src/pyopencl_ext.h
```diff
@@ -49,6 +49,10 @@ typedef union
 #define CL_DEVICE_PCI_SLOT_ID_NV                        0x4009
 #endif
 
+#ifndef CL_DEVICE_PCI_DOMAIN_ID_NV
+#define CL_DEVICE_PCI_DOMAIN_ID_NV                      0x4010
```
inducer, line 54:
```suggestion
/* https://github.com/pocl/pocl/pull/910 */
#define CL_DEVICE_PCI_DOMAIN_ID_NV                      0x4010
```

src/clinfo_ext.h
```diff
@@ -49,6 +49,7 @@
 #define CL_DEVICE_ATTRIBUTE_ASYNC_ENGINE_COUNT_NV	0x4007
 #define CL_DEVICE_PCI_BUS_ID_NV				0x4008
 #define CL_DEVICE_PCI_SLOT_ID_NV			0x4009
+#define CL_DEVICE_PCI_DOMAIN_ID_NV          0x4010
```
inducer, line 53:
```suggestion
/* https://github.com/pocl/pocl/pull/910 */
#define CL_DEVICE_PCI_DOMAIN_ID_NV          0x4010
```

doc/make_constants.py
inducer: Copy-paste error. Could you fix that?

matthiasdiener: Fixed here and at #425.

inducer:
Thanks!

Oblomov:
Sorry for the late comment, but is this officially part of the extension? My CUDA driver doesn't seem to accept this particular query.

matthiasdiener:
> Sorry for the late comment, but is this officially part of the extension? My CUDA driver doesn't seem to accept this particular query.

I don't think this is officially part of the extension, but it seems that the other IDs are also not part of it (see e.g., https://www.khronos.org/registry/OpenCL/extensions/nv/cl_nv_device_attribute_query.txt which does not mention the BUS_ID or SLOT_ID).

But you are right, it seems that I implemented the wrong number for DOMAIN_ID. Can you try to query ID `0x400A`? You can check if the retuned value matches the actual DOMAIN_ID via `nvidia-smi`. `0x400A` works on my system with the Nvidia driver.

Oblomov:
Hello @matthiasdiener , I'm aware that not all queries are publicly documented, but 0x4010 gave an error on my NVIDIA driver. I have tested with `0x400A` and the NVIDIA driver does return 0 (which is the only domain ID on my system). Good to know, I'll add it to `clinfo` too.

matthiasdiener:
> Hello @matthiasdiener , I'm aware that not all queries are publicly documented, but 0x4010 gave an error on my NVIDIA driver. I have tested with `0x400A` and the NVIDIA driver does return 0 (which is the only domain ID on my system). Good to know, I'll add it to `clinfo` too.

Thanks! I'll submit a PR to change pyopencl to use `0x400A` for the DOMAIN_ID.

inducer:
We should probably update pocl, too.

matthiasdiener:
> We should probably update pocl, too.

Patch is here: https://github.com/pocl/pocl/pull/917

PR #429, opened 2021-02-17
Mempool size stats

inducer:
@matthiasdiener Could you review?

### DIFF
--- a/doc/tools.rst
+++ b/doc/tools.rst
@@ -134,6 +134,20 @@ not complicated::
         The number of blocks in active use that have been allocated
         through this pool.
 
+    .. attribute:: managed_bytes
+
+        "Managed" memory is "active" and "held" memory.
+
+        .. versionadded: 2021.1.2
+
+    .. attribute:: active_bytes
+
+        "Active" bytes are bytes under the control of the application.
+        This may be smaller than the actual allocated size reflected
+        in :attr:`managed_bytes`.
+
+        .. versionadded: 2021.1.2
+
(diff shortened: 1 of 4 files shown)


src/mempool.hpp
```diff
@@ -291,12 +306,18 @@ namespace PYGPU_PACKAGE
         free_held();
       }
 
-      unsigned active_blocks()
+      size_type active_blocks()
```
matthiasdiener: Could probably mark these methods `const`, but not very important.

doc/tools.rst
```diff
@@ -134,6 +134,20 @@ not complicated::
         The number of blocks in active use that have been allocated
         through this pool.
 
+    .. attribute:: managed_bytes
```
matthiasdiener, line 137: Imho, it's a bit confusing that it's `held_blocks`, but not `held_bytes`, but doesn't matter much I think.

matthiasdiener (approved):

doc/tools.rst
inducer: I agree. However `held_bytes` is harder to track (/needs to be updated in more places) than these two, so I chose managed/active as the internal state. And computing `held_bytes` as `managed - active` is not correct, since active allocations are counted with their application-requested size, not their allocation size.

src/mempool.hpp
inducer: eba3dcc

inducer:
@matthiasdiener Thanks for taking a look!

PR #433, opened 2021-02-22
Fix dead link

James-E-A:
It appears Dr. Dobbs has done some spring-cleaning on his blog.

I chose that particular snapshot because it's adjacent to the last [snapshot of Part 2](http://web.archive.org/web/20190718032812/www.drdobbs.com/open-source/easy-opencl-with-python/240162614?pgno=2), to avoid the Next link accidentally glomming onto an error page in the future.


inducer:
Thanks!

PR #436, opened 2021-03-06
Add examples

numa65536:
Add 3 programs oftenly used un Blaise Pascal Center to compare all CPU and GPU, from Tesla C1060 to A100, from Radeon 7970 to RX6900XT.
PiXPU.py : estimation of Pi with Monte Carlo method, distribution of iterations over all processing units. Coarse Grain program
NBody.py : N-Body simulation with possible OpenGL rendering of set of particules using Newton second law.
TrouNoir.py : computing of image of accretion disk around Black Hole.

### DIFF
--- a/examples/NBody.py
+++ b/examples/NBody.py
@@ -1,0 +1,921 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+NBody Demonstrator implemented in OpenCL, rendering OpenGL
+
+By default, rendering in OpenGL is disabled. Add -g option to activate.
+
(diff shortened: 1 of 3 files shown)


numa65536:
Remove default export image under matplotlib. Revert to PIL

numa65536:
These 3 programs are a part of test and matrix programs from bench4gpu project. : https://forge.cbp.ens-lyon.fr/redmine/projects/bench4gpu
Two of the three examples above can be also launch with PyCUDA. A user can compare easily the difference between implementing a code in PyOpenCL and PyCUDA.

inducer:
Thanks! These will be valuable for users at many levels to see how the package can be used.

PR #441, opened 2021-03-15
move directory creation inside the wheel context

### LINKED ISSUE #438: Wheel creation creates .libs file which breaks installation of binary wheels

drewbo:
actually close #438 cc: @isuruf thanks for the tip and apologize for the double PR

### DIFF
--- a/scripts/fix-wheel.py
+++ b/scripts/fix-wheel.py
@@ -7,11 +7,11 @@ from delocate import wheeltools
 
 def add_library(paths):
     wheel_fnames = glob('/io/wheelhouse/pyopencl*.whl')
-    libs_dir = os.path.join('pyopencl', '.libs')
-    os.makedirs(libs_dir, exist_ok=True)
     for fname in wheel_fnames:
         print('Processing', fname)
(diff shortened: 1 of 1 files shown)


inducer:
Could you enable maintainer pushes? Alternatively, merge this up to main.

inducer:
Thanks!

inducer:
2021.1.4 released, hot off the presses just for you. Wheel builds should show up soon. Let me know if you continue to have trouble.

drewbo:
Thanks! The package installs correctly now but runs into the same error seen here: https://github.com/inducer/pyopencl/issues/391#issuecomment-717515393, I'll continue to investigate and open another issue/PR if necessary

PR #455, opened 2021-04-02
Drop dependency on decorator module

inducer:
See https://github.com/illinois-ceesd/mirgecom/issues/302 for context.


inducer:
Downstream boxtree CI is being held up by https://gitlab.tiker.net/inducer/boxtree/-/merge_requests/102 aka https://github.com/inducer/boxtree/pull/30.

PR #456, opened 2021-04-04
Implement pyopencl.array.stack

kaushikcfd:


pyopencl/array.py
```diff
@@ -2579,6 +2579,38 @@ def hstack(arrays, queue=None):
...
     return result
 
+
+def stack(arrays, axis=0, queue=None):
+    if not arrays:
+        raise ValueError("need at least one array to stack")
+
+    if axis != 0:
+        # pyopencl.Array.__setitem__ does not support non-contiguous assignments
+        raise NotImplementedError("axis!=0 not implemented")
```
inducer: This is only true for C-layout arrays. (`ary.flags.c_contiguous`) So IMO this needs to be checked.

inducer: You can easily generalize this to Fortran-layout arrays `ary.flags.f_contiguous` by transposing inputs and outputs. (should test this as well)

pyopencl/array.py
```diff
@@ -2579,6 +2579,38 @@ def hstack(arrays, queue=None):
 
     return result
 
+
+def stack(arrays, axis=0, queue=None):
```
inducer, line 2583: Add to docs.

inducer (commented):
Thanks! Just a few nits to pick, good once those are done.

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

pyopencl/array.py
kaushikcfd: f7c6bff226a55c1e4b95cf71bd819ffb9fe1c8f9

pyopencl/array.py
kaushikcfd: Yep, works for f_contiguous arrays too: ad6b6957a0fef0e65b55660311d5830a2c72e8b5

kaushikcfd: Yep, wasn't correct for all array layout orderings. Fixed in ad6b6957a0fef0e65b55660311d5830a2c72e8b5.

kaushikcfd:
@inducer: Thanks, this is ready for another review!

inducer:
Thx!

PR #457, opened 2021-04-05
Extend cl.array.(maximum|minimum|if_positive) to take scalars

kaushikcfd:


pyopencl/array.py
```diff
@@ -42,6 +42,9 @@
         get_common_dtype as _get_common_dtype_base)
 from pyopencl.characterize import has_double_support
 from pyopencl import cltypes
+from numbers import Number
+
+SCALAR_CLASSES = (Number, np.number, np.bool_, bool)
```
inducer: Doesn't `isinstance(x, np.number)` imply `isinstance(x, Number)`?

inducer: Add a test?

pyopencl/array.py
```diff
@@ -2673,6 +2676,17 @@ def if_positive(criterion, then_, else_, out=None, queue=None):
...
     """
 
+    if all(isinstance(k, SCALAR_CLASSES) for k in [criterion,
+                                                   then_,
+                                                   else_]):
+        result = np.where(criterion, then_, else_)
+
+        if out is not None:
+            out[...] = result
+            return
```
inducer: The non-scalar code path unconditionally returns `out`.

pyopencl/array.py
```diff
@@ -2673,6 +2676,17 @@ def if_positive(criterion, then_, else_, out=None, queue=None):
     contains *then_[i]* if *criterion[i]>0*, else *else_[i]*.
     """
 
+    if all(isinstance(k, SCALAR_CLASSES) for k in [criterion,
+                                                   then_,
+                                                   else_]):
```
inducer: This'll probably be quicker to execute as `isinstance() and isinstance() and isinstance()` (no separate stack frame setup, for instance).

inducer (commented):
Thanks! A few nits to pick below.

pyopencl/array.py
kaushikcfd: Yep removed the redundancy here. Thanks! 21fa507

kaushikcfd: Yep, thanks! 06c1e0d

pyopencl/array.py
kaushikcfd: Whoops fixed in <del>f706943</del>991135bfddf304b082b790912a7e76c65f48d176.

pyopencl/array.py
kaushikcfd: <del>9e9571f</del> eb29e2911c7c846fd9e1610df80d49459d91ea01

kaushikcfd:
@inducer: Thanks a lot for the suggestions. I believe this is ready.

inducer:
Thanks!

PR #470, opened 2021-05-13
Examples cleanup

inducer:


inducer:
FYI @numa65536 This contains updated versions of your examples that are reformatted to pass the Flake8 linter.

PR #485, opened 2021-06-14
cl.array.Array.ravel: takes order as an argument

kaushikcfd:
Numpy's [ravel](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html) takes an order argument, so it makes sense that pyopencl's array also take one. It's quite convenient to flatten F-contiguous nd-arrays (n > 1).


inducer:
Thanks!

PR #495, opened 2021-06-26
Return NotImplented for arithmetic on non-scalars, non-cl.Arrays

kaushikcfd:
On `main`, the following (buggy) behavior is seen:
```python
$ cat fail_pyopencl_arith.py 
from dataclasses import dataclass
import numpy as np
import pyopencl as cl
import pyopencl.array as cla

cq = cl.CommandQueue(ctx)

@dataclass
class ArrayContainer:
    _data: np.ndarray

    def __eq__(self, other):
        return ArrayContainer(self._data == other)

print(ArrayContainer(np.ones(100)) + cla.zeros(cq, (10,), dtype=np.float64))

$ python fail_pyopencl_arith.py 
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
```

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -1067,7 +1067,7 @@ class Array:
                         other.dtype.type(1), other))
 
             return result
-        else:
+        elif np.isscalar(other):
             # add a scalar
             if other == 0:
                 return self.copy()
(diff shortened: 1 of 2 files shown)


inducer:
Yikes, this is bad. Thanks for catching and fixing this. I'll roll a release to avoid other folks being bitten by it.

inducer:
436a5cbd2e25c5ef0eff610df0a13a7a18803f45 / https://pypi.org/project/pyopencl/2021.2.4/

PR #504, opened 2021-07-15
Cache: reduce sleep time

matthiasdiener:
CC: #503.

### DIFF
--- a/pyopencl/cache.py
+++ b/pyopencl/cache.py
@@ -89,17 +89,24 @@ class CacheLockManager(CleanupBase):
                     pass
 
                 wait_time_seconds = 0.05
+
+                # Warn every 10 seconds if not able to acquire lock
+                warn_attempts = int(10/wait_time_seconds)
+
+                # Exit after 60 seconds if not able to acquire lock
+                exit_attempts = int(60/wait_time_seconds)
+
                 from time import sleep
                 sleep(wait_time_seconds)
 
                 attempts += 1
 
-                if attempts % (10/wait_time_seconds) == 0:
+                if attempts % warn_attempts == 0:
                     from warnings import warn
                     warn("could not obtain cache lock--delete '%s' if necessary"
                             % self.lock_file)
 
-                if attempts > 60 / wait_time_seconds:
+                if attempts > exit_attempts:
                     raise RuntimeError("waited more than one minute "
                             "on the lock file '%s'"
                             "--something is wrong" % self.lock_file)
--- a/pyopencl/cache.py
+++ b/pyopencl/cache.py
@@ -88,6 +88,12 @@ class CacheLockManager(CleanupBase):
                 except OSError:
                     pass
 
+                # This value was chosen based on the py-filelock package:
+                # https://github.com/tox-dev/py-filelock/blob/a6c8fabc4192fa7a4ae19b1875ee842ec5eb4f61/src/filelock/_api.py#L113
+                # When running pyopencl in an application with multiple ranks
+                # that share a cache_dir, higher timeouts can lead to
+                # application stalls even with low numbers of ranks.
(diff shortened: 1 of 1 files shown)


pyopencl/cache.py
```diff
@@ -98,8 +98,8 @@ def __init__(self, cleanup_m, cache_dir):
                     warn("could not obtain cache lock--delete '%s' if necessary"
```
inducer, line 112: This logic no longer makes sense. It'll now print this every 0.05 seconds.

matthiasdiener: Hmm, I'm not sure what to do here... Remove the warning completely, or print it just for `attempts == 20`?

inducer: It should be printed at reasonable intervals, every ten seconds or so, IMO.

matthiasdiener: Done in 8ab383a

pyopencl/cache.py
```diff
@@ -88,17 +88,18 @@ def __init__(self, cleanup_m, cache_dir):
...
 
+                wait_time_seconds = 0.05
                 from time import sleep
-                sleep(0.05)
+                sleep(wait_time_seconds)
 
                 attempts += 1
 
-                if attempts > 10:
+                if attempts % (10/wait_time_seconds) == 0:
```
inducer: That won't work reliably. `/` unconditionally has a floating point result, and comparing floating point numbers to zero is unreliable.

yxliang01:
Actually, what about don't do polling at all, i.e. no sleep. There's a lot more efficient system calls available like `inotify` and seemingly good quality python package `watchdog` (never use myself though) to do this cross-platform.

inducer:
Thanks for the suggestion. IMO, anytime there is contention for this lock, 
you're running your program poorly. (I.e. this should always be avoidable, 
e.g. by setting a private cache folder.) As such, I'm not very inclined to optimize that case.

pyopencl/cache.py
matthiasdiener: > That won't work reliably. `/` unconditionally has a floating point result, and comparing floating point numbers to zero is unreliable.

4d13b82a0c387901783ccf8c2b61eacaaddec7be should fix this

matthiasdiener:
> Thanks for the suggestion. IMO, anytime there is contention for this lock, you're running your program poorly. (I.e. this should always be avoidable, e.g. by setting a private cache folder.) As such, I'm not very inclined to optimize that case.

A bit of background on this PR: We keep seeing this timeout a lot when running multiple MPI ranks on a single node, which is a difficult situation to set private cache folders (especially when running interactively).

pyopencl/cache.py
```diff
@@ -88,18 +88,26 @@ def __init__(self, cleanup_m, cache_dir):
                 except OSError:
                     pass
 
+                wait_time_seconds = 0.05
```
inducer, line 97: What's the motivation for this change? It seems really short. If you were running this (by mistake, say) on 1000 nodes sharing a cache directory, I think the resulting thrashing would not be a good thing.

matthiasdiener: The specific value was taken from https://github.com/tox-dev/py-filelock/blob/a6c8fabc4192fa7a4ae19b1875ee842ec5eb4f61/src/filelock/_api.py#L113. I can't really predict what the result of this with thousands of nodes would be, but we already keep on hitting the 1 second timeout when running with 2 ranks on a single node, which I guess indicates that the existing polling time would cause minutes-long waiting times when running with thousands of ranks.

inducer: I can see that argument. Could you add a comment justifying this choice of time-out value? (making reference to the single-node interactive "test" use case)

matthiasdiener: Added in 387011e

Issue #506, opened 2021-08-02
Array.concatenate fails for axis!= 0


panchoop:
The function only works for axis = 0.

To replicate:
```
import pyopencl as cl
platform = cl.get_platforms()[0]
device = platform.get_devices()[0]
context = cl.create_some_context([device])
queue = cl.CommandQueue(context)

a = np.random.rand(2,2)
b = np.random.rand(2,2)

a_cl = clarray.to_device(queue, np.require(a, np.float64, 'C'))
b_cl = clarray.to_device(queue, np.require(b, np.float64, 'C'))

c_cl = clarray.concatenate( (a_cl, b_cl), axis=1)
````

[Error: "cannot assign between arrays of differing strides"](https://github.com/inducer/pyopencl/blob/dfc4e840cfa4b371a90c624e44bee5399b057681/pyopencl/array.py#L2117-L2118)

The error appears to be originated when [creating the buffer `results`,](https://github.com/inducer/pyopencl/blob/dfc4e840cfa4b371a90c624e44bee5399b057681/pyopencl/array.py#L2638) that is expected to contain the whole data.
Since it is initialized with shape (2, 2+2), its stride is different from both `a`, and `b`. 
Then `Array.setitem` fails as it is no yet implemented to set items with different stride.

I am unsure if it is a bug. If it is not yet implemented, maybe it could be handy to raise an error for axis != 0 inside concatenate.

inducer:
That's a limitation of the current implementation. There are not currently plans to address this. I'd be open to considering a PR that adds a better error message.

cc @kaushikcfd

PR #508, opened 2021-08-05
Added NotImplemented error to Array.concatenate

panchoop:
As suggested in [this issue](https://github.com/inducer/pyopencl/issues/506#issuecomment-892134405), a warning was added both in the code and in the documentation.

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -2605,8 +2605,8 @@ def multi_put(arrays, dest_indices, dest_shape=None, out=None, queue=None,
 def concatenate(arrays, axis=0, queue=None, allocator=None):
     """
     Return a :class:`Array` that is a concatenation of the input tuple of 
-    :class:`Array` along :arg axis:. **Warning** Only axis = 0 has been 
-    implemented.
+    :class:`Array` along :arg axis:. **Warning** The current implementation
+    allows only concatenation of row-major arrays along axis = 0.
 
     .. versionadded:: 2013.1
     """
@@ -2615,8 +2615,8 @@ def concatenate(arrays, axis=0, queue=None, allocator=None):
     shape = None
     if axis != 0:
         raise NotImplementedError("Axis != 0. "+
-                                  "To be implementedk when Array.setitems " +
-                                  "allows values with different stride".)
+                                  "To be implemented when Array.setitems " +
+                                  "allows values with different stride.")
 
 
     for i_ary, ary in enumerate(arrays):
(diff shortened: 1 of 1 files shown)


pyopencl/array.py
```diff
@@ -2604,11 +2604,20 @@ def make_func_for_chunk_size(chunk_size):
...
+    Return a :class:`Array` that is a concatenation of the input tuple of 
+    :class:`Array` along :arg axis:. **Warning** Only axis = 0 has been 
+    implemented.
+
     .. versionadded:: 2013.1
     """
     # {{{ find properties of result array
 
     shape = None
+    if axis != 0:
```
inducer, line 2616: This isn't correct. This should instead be based on contiguity. Column-major arrays for example allow concatenation along the last axis.

panchoop: I noticed that the code failed the tests because I missed a syntax error. Sorry for the noise. 

About contiguity, I did some tests, and the function does not work in any axis for colum-major arrays.[ In line 2571 of the Array.concatenate function](https://github.com/inducer/pyopencl/blob/dfc4e840cfa4b371a90c624e44bee5399b057681/pyopencl/array.py#L2638), the output array `result` is always a row-major array, leading to the same `NotImlementedError` due to miss-matching strides.

Simple example:
```
import pyopencl as cl
platform = cl.get_platforms()[0]
device = platform.get_devices()[0]
context = cl.create_some_context([device])
queue = cl.CommandQueue(context)

a = np.random.rand(2,2)
b = np.random.rand(2,2)

a_cl = clarray.to_device(queue, np.require(a, np.float64, 'F'))
b_cl = clarray.to_device(queue, np.require(b, np.float64, 'F'))

c_cl = clarray.concatenate( (a_cl, b_cl), axis=1)
```

It shouldn't be hard to allow column-major arrays, it is just a matter of initializing the result array with `order="F"` in these cases. The issue I see is that the Array class has no member that indicates the ordering. One could deduce the order by looking at the strides (if Array.strides[0] > Array.strides[-1]: row-major, else: column-major) but maybe it is too hacky. what do you think?

Another approach is the one taken in the [Array.stack function](https://github.com/inducer/pyopencl/blob/dfc4e840cfa4b371a90c624e44bee5399b057681/pyopencl/array.py#L2752), in which when the axis != 0, it is assumed that the input is column mayor. But it would also lead to unclear error messages.

inducer: Thanks for pointing out the issue with `concatenate` and column-major arrays, we should look into fixing that. I agree with your assessment about inspection of the strides, it's not quite elegant.

FWIW, CI is still not passing on this branch.

panchoop: Oh yeah! I forgot to push. It should work now.

PR #509, opened 2021-08-14
make _new_with_changes function also work in a subclass

haesleinhuepf:
Hi @inducer ,

I'd like to suggest a minor change that would allow us to subclass cl.array.Array conveniently. We had some [minor issue in a related project](https://github.com/clEsperanto/pyclesperanto_prototype/pull/131#issuecomment-898956354) and could solve it with that change suggested here.

Thanks for considering!

Best,
Robert

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -636,7 +636,7 @@ class Array:
         else:
             events = self.events
 
-        return Array(None, shape, dtype, allocator=allocator,
+        return type(self)(None, shape, dtype, allocator=allocator,
                 strides=strides, data=data, offset=offset,
                 events=events,
                 _fast=fast, _context=self.context, _queue=queue, _size=size)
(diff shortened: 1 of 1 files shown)


pyopencl/array.py
```diff
@@ -636,7 +636,7 @@ def _new_with_changes(self, data, offset, shape=None, dtype=None,
         else:
             events = self.events
 
-        return Array(None, shape, dtype, allocator=allocator,
+        return type(self)(None, shape, dtype, allocator=allocator,
```
```suggestion
        return self.__class__(None, shape, dtype, allocator=allocator,
```
Would this work, too? This can be a fairly hot code path, and if this helps avoid any function call overhead, it might be worth it. (I don't have data to back this up, but it *feels* faster :)

haesleinhuepf: Sure! Looks 👍

inducer:
Thanks!

PR #513, opened 2021-10-03
Extend if_positive to take scalars for then or else

alexfikl:
This allows `if_positive` to work with
* "host" scalars, mostly inspired by how `__pow__` works.
* "device" scalars, inspired by the other functions that broadcast over device scalars.


inducer:
LGTM, thanks!

inducer:
Released 2021.2.7 with these changes. Should be on conda forge tomorrow-ish.

PR #522, opened 2021-11-24
Fix small sphinx docs rendering issue

kandersolar:
On the [parallel algorithms](https://documen.tician.de/pyopencl/algorithm.html) docs page, there is an RST directive inside a literal block, which I assume must be unintentional:

image

Removing one of the trailing colons in the preceding line should allow the directive to be interpreted correctly instead of placed into a literal block, though I did not confirm this before opening the PR.


inducer:
LGTM. Thanks a lot for the fix!

PR #523, opened 2021-12-04
Make if_positive keep strides of then_ if known.

nchristensen:
The previous version of `if_positive` always returns a C-ordered array even if `then_` is not C-ordered. This breaks artificial viscosity in Mirgecom when an array context uses a Fortran style data layout. This change makes the output have the same strides as `then_` if the shape of `then_` is known, otherwise it defaults to use the same layout as `criterion`.

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -2864,7 +2864,7 @@ def if_positive(criterion, then_, else_, out=None, queue=None):
 
     if out is None:
 
-        if not then_.shape == ():
+        if then_.shape != ():
             out = empty_like(
                 then_, criterion.queue, allocator=criterion.allocator)
         else:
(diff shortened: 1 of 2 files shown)


pyopencl/array.py
```diff
@@ -2863,9 +2863,19 @@ def if_positive(criterion, then_, else_, out=None, queue=None):
         raise AssertionError()
 
     if out is None:
-        out = empty(
-                criterion.queue, criterion.shape, then_.dtype,
-                allocator=criterion.allocator)
+
+        if not then_.shape == ():
```
```suggestion
        if then_.shape != ():
```

inducer (commented):
LGTM, thanks! (I'll apply the suggestion.) I'm guessing you'd like me to roll a release?

nchristensen:
Yes please! Thanks!

inducer:
https://pypi.org/project/pyopencl/2021.2.10/

inducer:
(It should wind its way to conda forge before too long.)

PR #532, opened 2022-01-17
Allow unsigned int for fancy indexing

alexfikl:
Is there a reason this wasn't allowed?

Context: was trying to port `take` in `pytential` and an index was `uint8` (apparently because of [boxtree](https://github.com/inducer/boxtree/blob/bfc06e411679f512225a62e5fc74fb2489627016/boxtree/tree_build.py#L73)).

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -1988,7 +1988,7 @@ class Array:
         """
 
         if isinstance(index, Array):
-            if index.dtype.kind != "i":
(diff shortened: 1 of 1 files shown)


inducer:
I can't think of a reason why we shouldn't allow it. And, at any rate, there wasn't a comment saying not to allow unsigned. :)

inducer:
https://pypi.org/project/pyopencl/2021.2.13/

PR #533, opened 2022-02-10
Fix, test integer ary/ary division to match numpy

inducer:
cc @majosm


inducer:
https://pypi.org/project/pyopencl/2022.1/

PR #551, opened 2022-04-03
Better support subclasses in __repr__

alexfikl:
This was motivated by giving a clearer repr to `TaggableCLArray` in https://github.com/inducer/arraycontext.

A bit shifty because it says `cl.TaggableCLArray`, which might confuse people to think it's located in `pyopencl`. What do you think?

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -859,28 +859,29 @@ class Array:
 
     def __str__(self):
         if self.queue is None:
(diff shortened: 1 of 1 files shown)


inducer:
I think it's an improvement. I like it better than the full `{tp.__module__}.{tp.__name__}`. Let's try it and see if anybody hates it.

Thanks for making this!

PR #555, opened 2022-04-09
Warn when a kernel has many arguments

matthiasdiener:
cc: @inducer @kaushikcfd

### DIFF
--- a/pyopencl/invoker.py
+++ b/pyopencl/invoker.py
@@ -315,6 +315,12 @@ def generate_enqueue_and_set_args(function_name,
         arg_types,
         work_around_arg_count_bug, warn_about_arg_count_bug):
 
+    if num_cl_args > 500:
+        from warnings import warn
+        warn(f"Kernel '{function_name}' has {num_cl_args} arguments. "
+              "This might lead to compilation errors on CUDA devices, "
+              "which have a limit of ~4 KByte of kernel arguments.")
+
     cache_key = (function_name, num_passed_args, num_cl_args,
             arg_types, __debug__,
             work_around_arg_count_bug, warn_about_arg_count_bug)
--- a/pyopencl/invoker.py
+++ b/pyopencl/invoker.py
@@ -315,8 +315,8 @@ def _get_max_parameter_size(dev):
     dev_limit = dev.max_parameter_size
 
     if get_pocl_version(dev.platform) is not None:
-        # Older pocl versions have an incorrect parameter size limit of 1024
-        # See e.g. https://github.com/pocl/pocl/pull/1046
+        # Current pocl versions (as of 04/2022) have an incorrect parameter
+        # size limit of 1024; see e.g. https://github.com/pocl/pocl/pull/1046
         if dev_limit == 1024:
             if dev.type & cl.device_type.CPU:
                 return 1024*1024
@@ -337,14 +337,20 @@ def _check_arg_size(function_name, num_cl_args, arg_types, devs):
 
         total_arg_size = 0
 
+        is_estimate = False
+
         if arg_types:
             for arg_type in arg_types:
-                if arg_type is None or isinstance(arg_type, VectorArg):
+                if arg_type is None:
+                    is_estimate = True
+                    total_arg_size += dev_ptr_size
+                elif isinstance(arg_type, VectorArg):
                     total_arg_size += dev_ptr_size
                 else:
                     total_arg_size += np.dtype(arg_type).itemsize
         else:
             # Estimate that each argument has the size of a pointer on average
+            is_estimate = True
             total_arg_size = dev_ptr_size * num_cl_args
 
         if total_arg_size > dev_limit:
@@ -353,7 +359,7 @@ def _check_arg_size(function_name, num_cl_args, arg_types, devs):
                 f"a total size of {total_arg_size} bytes, which is higher than "
                 f"the limit of {dev_limit} bytes on {dev}. This might "
                 "lead to compilation errors, especially on GPU devices.")
-        elif total_arg_size >= dev_limit * 0.75:
+        elif is_estimate and total_arg_size >= dev_limit * 0.75:
             # Since total_arg_size is just an estimate, also warn in case we are
             # just below the actual limit.
             from warnings import warn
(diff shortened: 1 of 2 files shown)


pyopencl/invoker.py
```diff
@@ -315,6 +315,12 @@ def generate_enqueue_and_set_args(function_name,
         arg_types,
         work_around_arg_count_bug, warn_about_arg_count_bug):
 
+    if num_cl_args > 500:
```
inducer: [CL_​DEVICE_​MAX_​PARAMETER_​SIZE](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#CL_DEVICE_MAX_PARAMETER_SIZE)

matthiasdiener: https://github.com/pocl/pocl/pull/1046 ;-)

Is it ok to estimate the argument size based on `num_cl_args*8` or should I look at each argument's actual size (how can I do that for a VectorArg?)?

inducer: This is not a performance-sensitive bit of code, so I see no reason to not at least try to be precise.

>  (how can I do that for a VectorArg?)?

https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_ADDRESS_BITS

matthiasdiener: An issue I couldn't resolve: the invoker is device-independent, but [CL_​DEVICE_​MAX_​PARAMETER_​SIZE](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#CL_DEVICE_MAX_PARAMETER_SIZE) is not. How can we resolve this?

- Somehow pass the device to `generate_enqueue_and_set_args` (how?)
- `gen()` the argument checking code into the invoker (potential performance hit, also probably too late to meaningfully warn about the arg count)
- Add the argument checking code somewhere else (where?)

inducer: A kernel is associated with a context, and from a context you can find the devices.

matthiasdiener: Thanks! What do you think of 24d00a40250033f1ffedf09fbc6d727d1dd13c1b?

pyopencl/invoker.py
```diff
@@ -313,7 +313,42 @@ def gen_arg_setting(in_enqueue):
...
         arg_types,
-        work_around_arg_count_bug, warn_about_arg_count_bug):
+        work_around_arg_count_bug, warn_about_arg_count_bug, devs):
+
+    # {{{ Check whether argument size exceeds the OpenCL device limit
+
+    if arg_types:
+        for dev in devs:
+            arg_size = 0
+            dev_limit = dev.max_parameter_size
```
matthiasdiener: Note that this will currently cause a lot of warnings in mirgecom, at least until https://github.com/pocl/pocl/pull/1046 is merged.

pyopencl/invoker.py
```diff
@@ -313,7 +313,42 @@ def gen_arg_setting(in_enqueue):
...
+                warn(f"Kernel '{function_name}' has {num_cl_args} arguments with "
+                    f"a total size of {arg_size} bytes, which approaches "
+                    f"the limit of {dev_limit} bytes on {dev}. This might "
+                    "lead to compilation errors especially on CUDA devices.")
+    else:
+        if num_cl_args > 500:
+            from warnings import warn
+            warn(f"Kernel '{function_name}' has {num_cl_args} arguments. "
+                "This might lead to compilation errors on CUDA devices, "
+                "which have a limit of ~4 KByte of kernel arguments.")
```
inducer: You could make a bold assumption that each arg occupies `dev_ptr_size` on average, guessing that most args are pointers. That would make this less arbitrary.

pyopencl/invoker.py
inducer: You can include logic to catch "if pocl and GPU and max_parameter_size == 1024" to mitigate that.

pyopencl/invoker.py
matthiasdiener: Done in [5addf38](https://github.com/inducer/pyopencl/pull/555/commits/5addf3818f2d145a5cace90c6b2fc7e48e754350)

pyopencl/invoker.py
matthiasdiener: Done in [5addf38](https://github.com/inducer/pyopencl/pull/555/commits/5addf3818f2d145a5cace90c6b2fc7e48e754350)

matthiasdiener:
A couple of actual driver limits, as reported by [CL_​DEVICE_​MAX_​PARAMETER_​SIZE](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#CL_DEVICE_MAX_PARAMETER_SIZE):

- pocl 1.8:
  - pthreads: 1024 (but probably unbounded)
  - cuda: 1024 (4352 in reality)
- Nvidia CL: 4352
- Apple CL:
  - CPU: 4096
  - Intel GPU: 1024
  - AMD GPU: 1024
- Intel CL: 3840

pyopencl/invoker.py
```diff
@@ -305,6 +306,65 @@ def gen_arg_setting(in_enqueue):
...
+        else:
+            # Estimate that each argument has the size of a pointer on average
+            total_arg_size = dev_ptr_size * num_cl_args
+
+        if total_arg_size > dev_limit:
+            from warnings import warn
+            warn(f"Kernel '{function_name}' has {num_cl_args} arguments with "
+                f"a total size of {total_arg_size} bytes, which is higher than "
+                f"the limit of {dev_limit} bytes on {dev}. This might "
+                "lead to compilation errors especially on CUDA devices.")
```
inducer:
```suggestion
                "lead to compilation errors, especially on GPU devices.")
```

pyopencl/invoker.py
```diff
@@ -305,6 +306,65 @@ def gen_arg_setting(in_enqueue):
...
+        elif total_arg_size >= dev_limit * 0.75:
+            # Since total_arg_size is just an estimate, also warn in case we are
```
inducer, line 363: In some cases it's exact, in others it's an estimate. This should only trigger if we're estimating.

matthiasdiener: What do you think of [db9bef2](https://github.com/inducer/pyopencl/pull/555/commits/db9bef2b90e08a7fc64ddab68c93aeeb69570273)?

inducer: Yep, that's what I had in mind.

inducer:
LGTM, thanks!

PR #558, opened 2022-05-01
Fix fallback in has_struct_arg_count_bug

alexfikl:


pyopencl/characterize/__init__.py
```diff
@@ -378,7 +378,7 @@ def has_struct_arg_count_bug(dev, ctx=None):
     if dev.platform.name == "Apple" and dev.type & cl.device_type.CPU:
         return "apple"
     if dev.platform.name == "Portable Computing Language":
-        pocl_version = get_pocl_version(dev.platform, fallback_value=(0.14))
+        pocl_version = get_pocl_version(dev.platform, fallback_value=(0, 14))
```
alexfikl, line 381: This is the actual fix: it used `0.14` instead of `0, 14`

.github/workflows/ci.yml
```diff
@@ -176,7 +176,7 @@ jobs:
                 ./configure.py --cl-inc-dir="$TEST_ENV_ROOT/include" --cl-lib-dir="$TEST_ENV_ROOT/lib"
                 git add -f siteconf.py
 
-                prepare_downstream_build "$DOWNSTREAM_PROJECT"
+                prepare_downstream_build "https://github.com/inducer/$DOWNSTREAM_PROJECT.git"
```
inducer, line 179: How come the old version no longer works?

alexfikl: Because I didn't realize that `prepare_downstream_build` was actually used outside of `ci-support` and did the backwards compatibility stuff only in the `test_downstream` wrapper.

Does this break anywhere else? 

x-ref: https://github.com/inducer/ci-support/pull/22

inducer: Fair, I didn't catch that either. This works for me.

inducer:
Thanks!

PR #564, opened 2022-05-19
log large argument counts

matthiasdiener:
This helps with debugging how argument counts increase, even at a point where they are not yet close to the device limit.

### DIFF
--- a/pyopencl/invoker.py
+++ b/pyopencl/invoker.py
@@ -31,6 +31,9 @@ from pytools.py_codegen import Indentation, PythonCodeGenerator
 from pyopencl.tools import _NumpyTypesKeyBuilder, VectorArg
 import pyopencl as cl
 
+import logging
+logger = logging.getLogger(__name__)
+
 
 # {{{ arg packing helpers
 
@@ -367,6 +370,12 @@ def _check_arg_size(function_name, num_cl_args, arg_types, devs):
                 f"a total size of {total_arg_size} bytes, which approaches "
                 f"the limit of {dev_limit} bytes on {dev}. This might "
                 "lead to compilation errors, especially on GPU devices.")
+        elif num_cl_args > 50:
+            logger.info(
+                f"Kernel '{function_name}' has {num_cl_args} arguments with "
+                f"a {'total estimated' if is_estimate else 'total'} size of "
+                f"{total_arg_size} bytes.")
+
 
 # }}}
 


inducer:
I'm honestly not sure about this. I think there are legitimate (if rare) use cases for kernels with that many arguments, and I think it's kind of presumptuous to tell people how many arguments they should be using. :)

matthiasdiener:
> I'm honestly not sure about this. I think there are legitimate (if rare) use cases for kernels with that many arguments, and I think it's kind of presumptuous to tell people how many arguments they should be using. :)

Right, I agree. But we aren't emitting a warning, just logging the values. The limit in this PR (50) is high enough so it will emit few messages even for mirgecom (for reference, autoignition-mpi with nspecies=7 emits two of of those).

inducer:
If I was an outside user (and, heaven forbid, this package does have those) and I got noisy log nonsense about argument counts, I think I would not be happy.

matthiasdiener:
> If I was an outside user (and, heaven forbid, this package does have those) and I got noisy log nonsense about argument counts, I think I would not be happy.

Are there ways to reduce the noisiness? E.g. wrapping this with `__debug__`. 

 This feature was specifically requested for mirge development and it is not easy to add this functionality by hand (as pyopencl is kind of hard to build). We already have lots of other debug messages that are quite noisy imho (e.g. the long build time messages).

pyopencl/invoker.py
```diff
@@ -367,6 +370,12 @@ def _check_arg_size(function_name, num_cl_args, arg_types, devs):
                 f"a total size of {total_arg_size} bytes, which approaches "
                 f"the limit of {dev_limit} bytes on {dev}. This might "
                 "lead to compilation errors, especially on GPU devices.")
+        elif num_cl_args > 50:
+            logger.info(
```
inducer, line 374:
```suggestion
            logger.debug(
```

inducer:
> This feature was specifically requested for mirge development

Requested? By whom? And why is this actionable and worthwhile as a message? And why should we prioritize mirgecom's needs over the everyone else's?

> We already have lots of other debug messages that are quite noisy imho

I find that the ones that we do have well-justified. If you've got counterexamples, please list them.

> (e.g. the long build time messages).

For example those: A user might be wondering what their code is doing if a compile is taking a while. Those log messages (at the INFO level, so not shown by default) provide an explanation.

To be a bit more explicit, and to avoid spending further time on this: This isn't happening.

PR #568, opened 2022-06-02
fix opencl algorithm: exit this loop if `z^2>4`

t0b3:
fix opencl algorithm: exit this loop if `z^2>4`


inducer:
Thanks, this makes sense!

t0b3:
you are welcome :+1:

PR #569, opened 2022-06-02
Mandelbrot example: double precision

t0b3:
add double precision for 
* `maxiter`: ushort(16) -> uint(32)
* `output`: ushort(16) -> uint(32)
* `z`, `q`: complex64 -> complex128
* `real`, `nreal`, `imag`: float -> double

works fine here

### DIFF
--- a/examples/demo_mandelbrot.py
+++ b/examples/demo_mandelbrot.py
@@ -38,7 +38,7 @@ def calc_fractal_opencl(q, maxiter):
     ctx = cl.create_some_context()
     queue = cl.CommandQueue(ctx)
 
-    output = np.empty(q.shape, dtype=np.uint16)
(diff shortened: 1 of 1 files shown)


inducer:
Could you explain what the presumed benefit is? An immediate downside is that devices without high-performance double support (Nvidia...) will suffer a decrease in performance. As a result, maybe this should be an option?

t0b3:
intend is to 
* extend the demo and explain how high precision can be implemented
* enable this very example to dive into much lower zoom levels
code works fine here on budget radeon

t0b3:
performance doesn't decrease compared to 32bit
fyi this 64bit code performs fluent on 64bit cpus like
* 5$ entry level cloud vm cpu
* raspi4 cpu

PR #571, opened 2022-06-07
Use dataclass instead of pytools.Record

alexfikl:
The change in `reduction.py` has some actual logic changes (made `_ReductionInfo` frozen so had to move things around), but the rest should be cleanup.

Most of these classes seemed internal, so I assume they don't need some compatibility with `Record` (?).

### DIFF
--- a/pyopencl/algorithm.py
+++ b/pyopencl/algorithm.py
@@ -29,13 +29,19 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 OTHER DEALINGS IN THE SOFTWARE.
 """
 
+from dataclasses import dataclass
+from typing import Optional
 
 import numpy as np
+
 import pyopencl as cl
-import pyopencl.array  # noqa
-from pyopencl.scan import ScanTemplate
+import pyopencl.array
+
+from pyopencl.elementwise import ElementwiseKernel
+from pyopencl.scan import ScanTemplate, GenericScanKernel
 from pyopencl.tools import dtype_to_ctype, get_arg_offset_adjuster_code
-from pytools import memoize, memoize_method, Record
+
+from pytools import memoize, memoize_method
 from mako.template import Template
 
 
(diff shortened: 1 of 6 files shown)


inducer:
Thanks! I'm open to a mutually dependent change in boxtree to help us get rid of `Record`.

alexfikl:
> Thanks! I'm open to a mutually dependent change in boxtree to help us get rid of `Record`.

I think I just missed a member in `BuiltList`, so hopefully the latest version will fix `boxtree`.

pyopencl/algorithm.py
```diff
@@ -29,13 +29,19 @@
 OTHER DEALINGS IN THE SOFTWARE.
 """
 
+from dataclasses import dataclass
```
inducer, line 32: The compat target for pyopencl is nominally still Py3.6, so this needs a conditional dependency in `setup.py`.

alexfikl: Right! Added in [f9e31f4](https://github.com/inducer/pyopencl/pull/571/commits/f9e31f41439c9a7fdfc0ae8476a05f4d8388e125).

pyopencl/array.py
```diff
@@ -2304,20 +2305,19 @@ def arange(queue, *args, **kwargs):
...
-    class Info(Record):
-        pass
+    @dataclass
+    class _ArangeInfo:
+        start: Optional[int] = None
+        stop: Optional[int] = None
+        step: Optional[int] = None
+        dtype: Optional["np.dtype"] = None
+        allocator: Optional[Any] = None
+        wait_for: List[cl.Event] = field(default_factory=list)
```
inducer: I think this might be better as a class-local.

inducer (approved):
Thanks! Just one request, otherwise this LGTM!

pyopencl/array.py
alexfikl: Function-local? Just `wait_for`, right?

inducer: What I mean is that I would prefer to not repeatedly incur the cost for code generation in `@dataclass`.

alexfikl: Ah, completely misunderstood that, sorry!

Moved it out of the function in https://github.com/inducer/pyopencl/compare/f9e31f41439c9a7fdfc0ae8476a05f4d8388e125..b0a975c4433fff5659fe55d8da3f117788bf199b.

inducer:
Thx!

PR #582, opened 2022-07-02
Inherit input array types in cl.array

alexfikl:
Fixes `maximum`, `minimum` and some other functions to return the same type as the input arrays.

For context: there were a couple of places in `arraycontext` that were still returning `cl.Array` instead of `TaggableCLArray`.

- [x] Needs inducer/boxtree#60

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -477,8 +477,8 @@ class Array:
                 queue = None
 
             else:
-                raise TypeError("cq may be a queue or a context, not '%s'"
-                        % type(cq))
+                raise TypeError(
+                    f"cq may be a queue or a context, not '{type(cq).__name__}'")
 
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! For a moment I was worried that dropping the tags would be worrisome, but that's actually desired behavior. Correct, @kaushikcfd?

kaushikcfd:
> I was worried that dropping the tags would be worrisome, but that's actually desired behavior.

Yep! Not propagating the tags for [TaggableCLArray](https://documen.tician.de/arraycontext/array_context.html?highlight=taggableclarray#arraycontext.impl.pyopencl.taggable_cl_array.TaggableCLArray) is the desired behavior.

inducer:
Great, thanks! Then this is good to go.

PR #586, opened 2022-07-19
fix compilation error with pybind 2.10

matthiasdiener:
Fixes the compilation error in CI due to (I suspect) https://github.com/pybind/pybind11/pull/4008


matthiasdiener:
This is ready for review @inducer

inducer:
Thank you!

PR #588, opened 2022-07-23
add scalar support to cla.min and cla.max

alexfikl:
Most other operations support scalars in some way or another. This should also match `numpy` now.
```
>> r = np.min(3)
>>> r
3
>>> type(r)
<class 'numpy.int64'>
```


inducer:
I feel like there's some sort of slippery slope here, but hey, let's ride along for a bit and see what happens. :)

PR #584, opened 2022-07-14
Fix `nan` handling in `min`/`max`/`minimum`/`maximum`

majosm:
Closes #583.

To-do:

- [x] Add failing tests
- [x] Fix the issue

PR #592, opened 2022-08-02
Implement dedicated kernels for `{min,max}imum`

inducer:
@majosm Turns out #584 mishandled infinities, which are used frequently as neutral elements.
@matthiasdiener This may be behind the NaNs that we were seeing, i.e. unrelated to the memory pool.

### DIFF
--- a/test/test_array.py
+++ b/test/test_array.py
@@ -1847,12 +1847,31 @@ def test_branch_operations_on_pure_scalars():
     cl_array.maximum,
     cl_array.minimum,
 ])
-def test_branch_operations_on_nans(ctx_factory, op):
+@pytest.mark.parametrize("special_a", [
+    np.nan,
+    np.inf,
+    -np.inf,
+])
+@pytest.mark.parametrize("special_b", [
+    np.nan,
+    np.inf,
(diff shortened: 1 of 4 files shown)


matthiasdiener:
> @matthiasdiener This may be behind the NaNs that we were seeing, i.e. unrelated to the memory pool.

Yes, this PR does fix the NaNs we are seeing.

test/test_array.py
```diff
@@ -1847,12 +1847,29 @@ def test_branch_operations_on_pure_scalars():
     cl_array.maximum,
     cl_array.minimum,
 ])
-def test_branch_operations_on_nans(ctx_factory, op):
+@pytest.mark.parametrize("special_a", [
+    np.nan,
+    np.inf,
```
matthiasdiener, line 1852: Does it make sense to add `-np.inf` here?

inducer: e8c0d023fa0a592b68de52dc709f2f48ed9e4129

inducer:
Force-merging since earlier tests passed and the latest changes only affected the (project-local) tests, which have passed.

PR #593, opened 2022-08-05
Use np.random.default_rng with a seed consistently

alexfikl:
Had this lying around from a while back. It changes to the new-ish `numpy` API for RNGs with seeds everywhere.

### DIFF
--- a/examples/ipython-demo.ipynb
+++ b/examples/ipython-demo.ipynb
@@ -1,202 +1,203 @@
 {
- "metadata": {
-  "name": "",
-  "signature": "sha256:81f3deed7cdc26b0fc756b3ee1eb6e8f9b1be96304ddfc6ff484d223c2b8a942"
- },
- "nbformat": 3,
- "nbformat_minor": 0,
- "worksheets": [
+ "cells": [
   {
-   "cells": [
-    {
-     "cell_type": "code",
-     "collapsed": false,
-     "input": [
-      "from __future__ import division\n",
-      "import numpy as np\n",
-      "import pyopencl as cl\n",
-      "import pyopencl.array"
-     ],
-     "language": "python",
-     "metadata": {},
-     "outputs": [
-      {
-       "output_type": "stream",
-       "stream": "stderr",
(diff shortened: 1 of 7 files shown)


examples/ipython-demo.ipynb
```diff
@@ -1,202 +1,203 @@
 {
- "metadata": {
-  "name": "",
-  "signature": "sha256:81f3deed7cdc26b0fc756b3ee1eb6e8f9b1be96304ddfc6ff484d223c2b8a942"
- },
- "nbformat": 3,
- "nbformat_minor": 0,
- "worksheets": [
+ "cells": [
```
alexfikl, line 2: Modified a single `np.random.rand` here and it updated the whole notebook :confounded: Let me know if you want this reverted or something?

examples/demo.py
```diff
@@ -3,8 +3,9 @@
 import numpy as np
 import pyopencl as cl
 
-a_np = np.random.rand(50000).astype(np.float32)
-b_np = np.random.rand(50000).astype(np.float32)
+rng = np.random.default_rng(seed=42)
+a_np = rng.random(50000, dtype=np.float32)
+b_np = rng.random(50000, dtype=np.float32)
```
inducer: The demos optimize for line count more than reproducibility, so I'd prefer if those got reverted to their prior state.

inducer (commented):
Thanks! Good to go once the demos are reverted.

examples/demo.py
alexfikl: Backed out all the changes in the demos in https://github.com/inducer/pyopencl/compare/8811123351a421cefc917f6a247190aea56558e9..7ca6985d92ce2e8ac42a34752e68831420b19eaf.

alexfikl:
@inducer Should be ready for another look. `ipython-demo.ipynb` has a lot of spurious changes because `jupyterlab` decided to convert it to "format v4", which seemed like a good idea to keep in.

inducer:
Thanks!

PR #452, opened 2021-03-29
Make SVM practical

inducer:
Draft because it's missing:

- [x] Tests
    - [x] ~~Ensure that SVM allocators can't be passed to buffer mem pool and vice versa.~~
- [x] A memory pool for SVM
    - [x] Quite subtle: #447 
- [x] Documentation, especially about synchronization of deallocation with pending work.
- [x] ~~`demo_array_svm` crashes for the moment,~~ the queue-associated SVM allocation is one potential solution for that, at least for in-order queues. See also #449.
- [x] via @matthiasdiener Check for needed virtual destructors via `clang`.
- [ ] Pass gitlab CI: https://gitlab.tiker.net/inducer/pyopencl/-/tree/array-svm
- [x] Do a full review

PR #594, opened 2022-08-08
simplify demo_array_svm and add SVM tests

matthiasdiener:
Merging into #452


inducer:
Thanks! I've incorporated these changes in #452.

PR #546, opened 2022-03-29
Switch from pybind11 to nanobind

inducer:
Surprisingly, this is mostly working now. :)

To do/Needs:
- [x] ~~Pass location of numpy headers to CMake~~ (found via CMake now)
- [x] ~~Pass OpenCL config info to CMake~~ (found via CMake now)
- [x] Actually fix the wrapper :slightly_smiling_face: 
    - [x] ` pycl test_wrapper.py 'test_mempool(cl._csc)' ` crashes
    - [x] `PYOPENCL_TEST=port python test_array.py "test_mem_pool_with_arrays(cl._csc)" ` crashes
    - [x] Doesn't make it through the meshmode test suite
- [x] https://github.com/wjakob/nanobind/pull/38
- [x] ~~https://github.com/wjakob/nanobind/pull/23~~
- [x] ~~https://github.com/wjakob/nanobind/pull/24~~
- [x] ~~See if the `_enqueue` trampoline can be avoided~~ (doesn't matter for loopy enqueues, that goes straight to `enqueue_nd_range_kernel`)
- [x] ~~https://github.com/wjakob/nanobind/pull/71~~
- [x] Expose SVM-as-buffer
- [x] Investigate reference leaks after tests: they've disappeared.
- [x] ~~Wait for nanobind 2, for `for_setter`.~~
- [x] Pass CI
- [x] Check loopy CI. Currently, it requires weak references to `cl.Context`, which nanobind doesn't currently allow. (Edit: [Now does allow.](https://nanobind.readthedocs.io/en/latest/classes.html#weak-references))
- [x] ~~Work around https://github.com/wjakob/nanobind/pull/72~~ (This seems to have been resolved in the meantime.)

PR #596, opened 2022-08-09
setup.py: add headers as dependencies

matthiasdiener:
Without this, modifications to header files do not trigger a
rebuild.

See also https://docs.python.org/3/distutils/setupscript.html#other-options

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -222,6 +222,15 @@ def main():
                     extra_compile_args=conf["CXXFLAGS"],
                     extra_link_args=conf["LDFLAGS"],
                     language="c++",
+                    depends=[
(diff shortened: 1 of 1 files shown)


inducer:
Sorry, no. distutils is not smart enough to gather dependencies in the way a reasonable build tool might. In addition, this is a half-measure: There are many other headers that really ought to be watched for changes.

#546 uses cmake to build, which solves this without having to clutter up `setup.py`.

inducer:
To further explain: The only reliable way to build of which I'm aware is to start from scratch every time, i.e.
```
rm -Rf build; python setup.py develop
```

PR #603, opened 2022-09-08
SVM: make sure to pop memcpy kwargs even with -O

matthiasdiener:
This fixes mirgecom SVM execution with -O

### DIFF
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -1971,9 +1971,9 @@ def enqueue_copy(queue, dest, src, **kwargs):
         # These are NOT documented. They only support consistency with the
         # Buffer-based API for the sake of the Array.
         if kwargs.pop("src_offset", 0) != 0:
-            raise AssertionError
+            raise ValueError("src_offset must be 0")
         if kwargs.pop("dst_offset", 0) != 0:
-            raise AssertionError
+            raise ValueError("dst_offset must be 0")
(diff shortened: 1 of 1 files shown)


pyopencl/__init__.py
```diff
@@ -1970,8 +1970,10 @@ def enqueue_copy(queue, dest, src, **kwargs):
 
         # These are NOT documented. They only support consistency with the
         # Buffer-based API for the sake of the Array.
-        assert kwargs.pop("src_offset", 0) == 0
-        assert kwargs.pop("dst_offset", 0) == 0
+        if kwargs.pop("src_offset", 0) != 0:
+            raise AssertionError
+        if kwargs.pop("dst_offset", 0) != 0:
+            raise AssertionError
```
inducer: It'd be better if (all of) these were `ValueError`s with an error message.

inducer (commented):
Yikes, good catch! Thanks for the patch. Just one minor wrinkle. Once this is in, I'll do another release.

pyopencl/__init__.py
matthiasdiener: What do you think of fd856a5?

inducer: That works, thanks!

inducer:
https://pypi.org/project/pyopencl/2022.2.1/

PR #604, opened 2022-09-08
Rework `_get_common_dtype` to tolerate unsized scalars, match numpy, warn about downcasts

matthiasdiener:
Otherwise, (e.g.) `cla.minimum(scalar, array)` won't work since scalar doesn't have a dtype attribute.

Fixes mirgecom execution with limiter.

Related to #592. This is a regression from 2022.1.6. The attached tests fail on current main.

- [ ] Pass gitlab CI. https://gitlab.tiker.net/inducer/pyopencl/-/commits/sym-dtype

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -42,8 +42,7 @@ from pyopencl.compyte.array import (
         f_contiguous_strides as _f_contiguous_strides,
         c_contiguous_strides as _c_contiguous_strides,
         equal_strides as _equal_strides,
-        ArrayFlags as _ArrayFlags,
-        get_common_dtype as _get_common_dtype_base)
+        ArrayFlags as _ArrayFlags)
 from pyopencl.characterize import has_double_support
 from pyopencl import cltypes
 from numbers import Number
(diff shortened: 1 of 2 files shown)


MTCam:
Fixes illinois-ceesd/mirgecom#747

pyopencl/array.py
```diff
@@ -66,7 +66,12 @@ def _get_common_dtype(obj1, obj2, queue):
 
     dsupport = has_double_support(queue.device)
     cache_key = None
-    o1_dtype = obj1.dtype
+    try:
+        o1_dtype = obj1.dtype
```
inducer: How about 

`getattr(ob1, "dtype", type(obj1))`

matthiasdiener: Done in [3dd634e](https://github.com/inducer/pyopencl/pull/604/commits/3dd634e609a2cd69ca5e58c2177eb2b06f1ec5d1)

matthiasdiener: Now I remember why I exchanged the arguments previously, compyte's `get_common_dtype` also doesn't accept a simple type as its first argument...

Should I modify compyte to support this or go back to my earlier implementation in this PR?

inducer: Try and modify compyte to see whether that works out.

matthiasdiener: > Try and modify compyte to see whether that works out.

With https://github.com/inducer/compyte/pull/40, this PR appears to pass the tests.

inducer:
I've reworked this a bit, and I'm inclined to merge it soon (tonight-ish) and release, assuming it passes CI. If you're around, I'd like it if you could give it a look.

inducer:
https://pypi.org/project/pyopencl/2022.2.2/

matthiasdiener:
I did a quick test of the reworked version, and it appears to fix https://github.com/illinois-ceesd/mirgecom/issues/747.

pyopencl/array.py
```diff
@@ -59,35 +58,77 @@
...
+        o2_dtype = np.array(obj2).dtype
+
+        # See above.
+        o2_is_array = True
+
+    cache_key = (o1_dtype, o2_dtype, o1_is_array, o2_is_array, allow_double)
+    #try:
+    #    return _COMMON_DTYPE_CACHE[cache_key]
+    #except KeyError:
+    #    pass
```
alexfikl, line 109: @inducer Is this meant to be commented out?

inducer: :facepalm: Thanks for pointing this out. I'll fix this, worry about the incompatibility in boxtree that broke things, and roll a new release.

inducer: https://gitlab.tiker.net/inducer/pyopencl/-/jobs/457531/raw is the boxtree failure I mean, though now I'm no longer sure this is with current pyopencl? The pytential failure may be related...

alexfikl: Ah! I got that in the array context branch and thought it was my fault for moving things around carelessly.

Yeah, the pytential issue can also be some dtype thing tripping up a tolerance check, so hopefully not too hard to fix :D

alexfikl: If nothing else, `pytential` passes (locally) with this reverted..

inducer: How extraordinarily sketchy. There shouldn't be a functional difference between use-the-cache and not. :facepalm:

alexfikl: Ah, no, I meant with this whole changeset (went back to 78ee8f95), not just this commented out chunk.

inducer: #609 should mop this up.

inducer: https://github.com/inducer/pyopencl/releases/tag/v2022.2.3 is winding its way onto the package index.

Issue #303, opened 2019-10-21
`Array.events` spuriously synchronizes on read-after-read


inducer:
- [ ] There should be `Array.write_events` and `Array.read_events`, with `Array.events` aliasing `write_events` under a deprecation.
- [ ] Arguments to all the generated kernels (elementwise etc.) should declare whether they're writing and automatically add to the appropriate `events` list.
- [ ] Audit code calling into the kernels to avoid duplicate event insertion.

cc @alexfikl   @zachjweiner

zachjweiner:
If I'm recalling correctly, I think the idea was that once we can distinguish between read- and write-synchronization points, there will be no need to enforce the order of two kernels based on an array that they both (only) read. For example, the operations `y = 2 * x` and `y_h = 2 * x.get()` do not need to be ordered, but the first would add an event to `x.events` which the call to `get` would wait for. In principle they could execute asynchronously, but `Array`'s event tracking enforces this synchronization at the moment (because it has no way to know that neither execution modifies `x`).

Does this make sense?

inducer:
Thanks for explaining, I remember now. The current `Array.events` system needs to be redone to wait on reads before making in-place changes.

[!91](https://gitlab.tiker.net/inducer/pyopencl/merge_requests/91) adds synchronization for WAR in `.get()`, but the overall issue is more general, IMO.

This is really what [gl-19](https://gitlab.tiker.net/inducer/pyopencl/issues/19) was meant to capture. (which we'll have this issue subsume)

inducer:
@zachjweiner Could you remind me which way this issue went? According to my experiments, I believe `2*x` does not *currently* add the resulting event to `x.events` (as of a60b4ed), so there's is not currently spurious synchronization. But I could see an argument that there *should* be synchronization for `2*x`, to ensure that nobody modifies `x` before `2*x` is done reading. Was this the core of the issue?

zachjweiner:
> I believe `2*x` does not currently add the resulting event to `x.events`

I find the same - it seems events are only appended to `self` in [array.py](https://github.com/inducer/pyopencl/blob/master/pyopencl/array.py) and never `other`.

> But I could see an argument that there should be synchronization for `2*x`, to ensure that nobody modifies `x` before `2*x` is done reading. Was this the core of the issue?

Yes, I believe so. And I definitely believe the argument.

[...]

PR #611, opened 2022-09-16
Add events to ElementwiseKernel args

alexfikl:
This is what I was trying to ask today, i.e. if the input / output args of `ElementwiseKernel` get the event that gets generated. Currently they do not. Should they?

For context, this is about removing the manual event handling in `boxtree` for something like the `GappyCopyAndMapKernel`
https://github.com/inducer/boxtree/blob/529e8d62f365350a80c9f16a0b12b9fb2c1a140e/boxtree/tools.py#L453-L509
which has a bunch of arguments.

### DIFF
--- a/test/test_algorithm.py
+++ b/test/test_algorithm.py
@@ -59,7 +59,8 @@ def test_elwise_kernel(ctx_factory):
             "linear_combination")
 
     c_gpu = cl_array.empty_like(a_gpu)
-    lin_comb(5, a_gpu, 6, b_gpu, c_gpu)
+    evt = lin_comb(5, a_gpu, 6, b_gpu, c_gpu)
+    assert evt in c_gpu.events
 
     assert la.norm((c_gpu - (5 * a_gpu + 6 * b_gpu)).get()) < 1e-5
 


inducer:
Hmm, thanks for the reminder. #303 is related. One complexity here is that `ElementwiseKernel` (at least without additional help) has no way to find out which array is being written. Tthough, in keeping with #303, read-after-write and write-after-read are also real dependencies. Part of me feels that we should just unconditionally add events to everything (reads and writes), but then another part of me would like to start tackling #303 first, by separating read events and write events.

inducer:
You closed this, but didn't leave a comment. Could you explain?

alexfikl:
@inducer From your comment, it seemed like #303 is needed before anything useful can be done here. And adding events unconditionally could also be done in `boxtree` as a stopgap, so I didn't think this was needed.

Do you think it would be better to do something like that (maybe under a flag) for the various kernel classes in here?

inducer:
I'm not sure it makes sense to clutter up boxtree with event management code. I think the desired goal state, from my perspective, would be to have array events split into read and write, and for that to be an effective change, for arguments to `ElementwiseKernel` to declare whether they're reading or writing. At that point, boxtree would need to be updated to declare read/write, and we could call this done. I'll reopen this to track that change. Sound fair?

inducer:
Or... since this is a PR, maybe not. I'll reclose it and make an issue. Sorry about the back-and-forth.

inducer:
Even better. I'll add this as an item under #303.

alexfikl:
> Sound fair?

Yeah, that sounds like a good plan to me!

PR #616, opened 2022-10-02
Add read_events and write_events to Array

### LINKED ISSUE #303: `Array.events` spuriously synchronizes on read-after-read

alexfikl:
Takes care of the first point from #303.

* Adds a `read_events` and `write_events` to `Array`.
* Teaches all the functions in `cl.array` to use those and not wait on read events when they don't need to.
* Teaches all the functions in `cl.clmath` to play nice as well
* The rest of the code just does a straight port form `events -> write_events`. This still needs work.

TODO:
- [ ] Needs corresponding loopy change

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -38,7 +38,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     ClassVar,
-    Concatenate,
     Literal,
     ParamSpec,
     cast,
@@ -221,12 +220,12 @@ def _splay(
 
 def get_wait_for_events(
         *,
-        outputs: Iterable[Any], inputs: Iterable[Any]) -> list[cl.Event]:
+        outputs: Iterable[Any], inputs: Iterable[Any]) -> cl.WaitList:
     # NOTE:
     # * outputs need to wait for all reads and writes to finish
     # * inputs only need to wait on writes, but not reads
 
-    wait_for = []
+    wait_for: cl.WaitList = []
     for ary in outputs:
         if isinstance(ary, Array):
             wait_for.extend(ary.write_events)
@@ -259,12 +258,18 @@ P = ParamSpec("P")
 
 
 def elwise_kernel_runner(
-            kernel_getter: Callable[Concatenate[Array, P], cl.Kernel]
-        ) -> Callable[Concatenate[Array, P], cl.Event]:
+            kernel_getter: Callable[P, cl.Kernel]
+        ) -> Callable[P, cl.Event]:
     """Take a kernel getter of the same signature as the kernel
     and return a function that invokes that kernel.
 
-    Assumes that the zeroth entry in *args* is an :class:`Array`.
+    The returned callable always takes additional keyword arguments:
+
+    * ``wait_for``: a sequence of events the kernel should wait for.
+    * ``queue``: the queue to execute the kernel on.
+    * ``noutputs``: the number of arguments that are considered outputs, e.g. if
+        the kernel takes 3 positional arguments and the first two are outputs,
+        this should be set to "2" (see ``frexp``).
     """
     from functools import wraps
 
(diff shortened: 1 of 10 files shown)


alexfikl:
@inducer This should be in a non-horrible state now. Can you take a quick look to see if it's doing what you had in mind? 

From going through it, it seems like several things still need work
* `elwise_kernel_runner` could probably add to `write_events` for the first arg and to `read_events` for the rest of them, so we don't clutter all the rest of the code. Does that sound reasonable?
* It seems to me that something like an `C = AXPBY` would need wait for `read_events` on `C` and `write_events` on `A` and `B`. Does that sound right? (+ similar logic for the other elementwise kernels)

inducer:
> * `elwise_kernel_runner` could probably add to `write_events` for the first arg and to `read_events` for the rest of them, so we don't clutter all the rest of the code. Does that sound reasonable?

Does.

> * It seems to me that something like an `C = AXPBY` would need wait for `read_events` on `C` and `write_events` on `A` and `B`. Does that sound right? (+ similar logic for the other elementwise kernels)

The way I see it, it wouldn't have to wait on `read_events`, because read-after-read does not need synchronization. Or am I missing something?

alexfikl:
> The way I see it, it wouldn't have to wait on `read_events`, because read-after-read does not need synchronization. Or am I missing something?

In that example, `C` would be a write-after-read, right? That's why I was thinking that all the reads should finish before it gets modified.

inducer:
> * 
> 
> * It seems to me that something like an `C = AXPBY` would need wait for `read_events` on `C` and `write_events` on `A` and `B`. Does that sound right? (+ similar logic for the other elementwise kernels)

`C` would need to wait for `read_events` *and* `write_events`, unless you'd make the two subsets of each other.

alexfikl:
@inducer This should be ready for a look. It could probably use some more tests, but I'm not quite sure what we can do there. Any ideas?

pyopencl/invoker.py
```diff
@@ -138,7 +138,7 @@ def add_buf_arg(arg_idx, typechar, expr_str):
                 cl_arg_idx += 1
 
             if in_enqueue:
-                wait_for_parts .append(f"{arg_var}.events")
```
inducer, line 147: Bump invoker cache version

inducer, line 147: Check that invoker cache is disabled by env var

alexfikl:
@inducer This is the macos error I was worried about
https://github.com/inducer/pyopencl/actions/runs/3210604562/jobs/5248179538
but it seems to be happening on other branches as well, e.g.
https://github.com/inducer/pyopencl/actions/runs/3254926391/jobs/5343705664
so it's likely no biggie (?). With that in mind, I feel better about this not introducing any cool new bugs :D

pyopencl/invoker.py
alexfikl: Done in #637

pyopencl/array.py
```diff
@@ -1836,25 +1854,48 @@ def conj(self):
...
-        list.
+        return self.add_write_event(evt)
+
+    def add_write_event(self, evt: "cl.Event") -> None:
+        """Add *evt* to :attr:`write_events`. If :attr:`write_events` is too
+        long, this method may implicitly wait for a subset of :attr:`write_events`
+        and clear them from the list.
         """
-        n_wait = 4
+        assert evt not in self.write_events
```
alexfikl: This is probably a bad idea (same for `read_events`)!

alexfikl:
This is shelved for the time being. Main reason being that it adds a bunch more bookkeeping for no clear benefit.

PR #620, opened 2022-10-03
Check isscalar in arithmetic functions

alexfikl:
Mostly the last commit is useful, the rest just clean up some docs, so probably better viewed one-by-one.


inducer:
Thanks!

PR #628, opened 2022-10-06
Remove wait_for handling from arange

alexfikl:
As far as I can tell, `arange` would raise a `ValueError` if `wait_for` was passed in, so this is not needed.

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -28,7 +28,7 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 OTHER DEALINGS IN THE SOFTWARE.
 """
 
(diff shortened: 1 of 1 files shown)


inducer:
Could you explain? It sounds like you're trying to argue that this isn't a compatibility break, but I'm not sure where the `ValueError` would have come from.

alexfikl:
This chunk seems to handle the `kwargs` and `wait_for` isn't allowed, so there's no point for it to do be in the dataclass.
https://github.com/inducer/pyopencl/blob/9e87128cc7907979eb9c3bbeaf765305a8e94004/pyopencl/array.py#L2473-L2484

inducer:
Thanks! Got it, in it goes.

PR #629, opened 2022-10-06
Import builtins in cl.array

alexfikl:
Instead of using `_builtin_*`.


inducer:
Definitely nicer, thanks!

PR #639, opened 2022-10-17
Add mypy to CI

alexfikl:
Was starting to add some types in #616 and friends, so thought we could check for it a bit too.

I had to re-order the shape handling a bit in `Array.__init__` to convince mypy it's a tuple at the end of the day. Let me know if it still looks ok.


alexfikl:
Running on Gitlab as well at 
https://gitlab.tiker.net/fikl2/pyopencl/-/pipelines/350236

inducer:
Thanks!

PR #642, opened 2022-10-19
Misc cleanup in elementwise

alexfikl:
Mostly adds docs, type annotations and some f-strings (commits should be fairly self-contained).


inducer:
LGTM, thanks!

PR #648, opened 2022-10-30
Add types and clean up docs in reduction.py

alexfikl:
Similar to #642.

PR #650, opened 2022-10-31
Add types and cleanup docs in scan.py

alexfikl:
Similar to #648.

### DIFF
--- a/doc/algorithm.rst
+++ b/doc/algorithm.rst
@@ -129,29 +129,10 @@ Making Custom Scan Kernels
 
 .. autoclass:: GenericScanKernel
 
-    .. method:: __call__(*args, allocator=None, queue=None, size=None, wait_for=None)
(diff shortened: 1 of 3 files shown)


alexfikl:
For some reason the types don't render in the docs (at least locally). There are no errors from `sphinx`, so I'll have to do some more digging :\

Screenshot_20221031_094812

alexfikl:
Managed to figure out what was going on in there :\

For future reference (if nothing else) the issue was with `index_dtype: Any = <class 'numpy.int32'>`. Basically `sphinx` would to a `repr(np.int32)` to make the whole signature into a string and then later on do a `ast.parse` to add back hyperlinks. The parse would then throw a `SyntaxError` on that part and abort the whole thing :cry:

inducer:
Thanks!

PR #652, opened 2022-11-02
Add MemoryObjectHolder to docs

alexfikl:
This was used in `enqueue_fill` but wouldn't link. `sphinx` would throw a warning when ran with
```
make html SPHINXOPTS="-W --keep-going -n -Dautodoc_typehints=both"
```

`enqueue_fill` doesn't pick up the arguments from the function because it's defined as
```
.. autofunction:: enqueue_fill(queue, dest, src, **kwargs)
```
so not quite sure if this is wanted?

### DIFF
--- a/doc/misc.rst
+++ b/doc/misc.rst
@@ -853,6 +853,10 @@ Internal Types
 
     See :class:`pyopencl.MemoryMap`.
 
+.. class:: MemoryObjectHolder
+
+    See :class:`pyopencl.MemoryObjectHolder`.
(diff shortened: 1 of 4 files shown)


inducer:
Thanks! `MemoryObjectHolder` is an internal interface that shouldn't have "leaked" into the signature. I've picked up the first commit in #653 and fixed the signature there, too. That should supersede this PR, closing here. Reopen if it seems I missed something.

PR #655, opened 2022-11-07
Fix some dead links

alexfikl:
Found when getting paranoid in #650 using
```
make linkcheck SPHINXOPTS="-W --keep-going -n" 
```
Mostly fixes dead links, changes redirects and uses `https` hopefully everywhere. These are only links that show up in the docs, not ones that show up in comments and stuff.


alexfikl:
The failure seems to be just a timeout, so restarting the tests should hopefully fix it.

inducer:
Thanks!

PR #656, opened 2022-11-09
Fix double precision check

alexfikl:
`mypy` 0.990 started complaining about this. Not quite sure what the original intent was?


inducer:
Oh :facepalm: . Thanks!

PR #658, opened 2022-11-12
Some more doc fixes

alexfikl:
Small fixes:

* replace single backticks with double backticks.
* fixed the indentation in memory pool docs. Currently the first paragraph is rendered weirdly (https://documen.tician.de/pyopencl/tools.html#pyopencl.tools.SVMPool) because `_MEMPOOL_IFACE_DOCS` is not indented.


inducer:
Thanks!

PR #660, opened 2022-11-16
Fix buffer size computation

### LINKED ISSUE #659: Array allocation failure
I have a code which does a series of allocations using `pyopencl.array`.  
One of the allocation fails with the following error

```
/mnt/multipath-shares/scisoft/tomotools_env/integrator/ubuntu20.04/ppc64le/dev/lib/python3.8/site-packages/silx/opencl/processing.py in allocate_buffers(self, buffers, use_array)
    215                     for buf in buffers:
    216                         print("####", buf.name, buf.size, buf.dtype)
--> 217                         mem[buf.name] = pyopencl.array.empty(self.queue, buf.size, buf.dtype)
    218                 else:
    219                     for buf in buffers:

/mnt/multipath-shares/scisoft/tomotools_env/integrator/ubuntu20.04/ppc64le/dev/lib/python3.8/site-packages/pyopencl/array.py in __init__(***failed resolving arguments***)
    684                     print("alloc_nbytes", alloc_nbytes) # DEBUG
    685 
--> 686                     self.base_data = cl.Buffer(
    687                             context, cl.mem_flags.READ_WRITE, alloc_nbytes)
    688                 else:

TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. pyopencl._cl.Buffer(context: pyopencl._cl.Context, flags: int, size: int = 0, hostbuf: object = None)

Invoked with: <pyopencl.Context at 0x2ec44690 on <pyopencl.Device 'Tesla V100-SXM2-32GB' on 'Portable Computing Language' at 0x2e32ac38>>, 1, 4471.0
```

~Unfortunately it's difficult to create a MWE to reproduce the crash, not sure why (caching ?).~ 
See https://github.com/inducer/pyopencl/issues/659#issuecomment-1318177386  for a MWE reproducing the crash

After investigation, it seems that the cause is in `pyopencl.array`  which tries to allocate a buffer where `size` is passed as float.  
Consider the following part in `pyopencl/array.py`: 

```python
            size = 1
            for dim in shape:
                size *= dim
                if dim < 0:
                    raise ValueError(
                        f"negative dimensions are not allowed: {shape}")
```

Importantly, if `size` is a `uint64` and `dim` is an integer, the `*=` operation is promoting `size` to a float (!).

I'd suggest to use the following:
```python
    size = numpy.prod(shape)
    if size < 0:
         raise ValueError(f"negative dimensions are not allowed: {shape}")
```

The above code snippet fixes the problem in my case.

Tested on python 3.8, pocl driver, power9 architecture, Nvidia V100 GPU:
  - Succeeds with pyopencl 2022.2.4 release (on pypi or github tag), 
  - Fails with current main branch
  - Succeeds with the above patch

pierrepaleo:
Fix #659

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -612,14 +612,16 @@ class Array:
 
                 shape = (shape,)
 
-            # Previously the size was computed as "size = 1; size *= dim for dim in shape
-            # However it can fail when using data types, eg., numpy.uint64(1) * 2  returns 2.0 !
+            # Previously the size was computed as
+            #   "size = 1; size *= dim for dim in shape"
+            # However it can fail when using data types,
+            # eg numpy.uint64(1) * 2  returns 2.0 !
             if np.any(np.array(shape) < 0):
                 raise ValueError(f"negative dimensions are not allowed: {shape}")
             if any([np.array([s]).dtype.kind not in ["u", "i"] for s in shp]):
                 raise ValueError(
-                "Invalid shape %s ; dimensions, must be integer" % (str(shape))
-            )
+                    "Invalid shape %s ; dimensions, must be integer" % (str(shape))
+                )
             size = np.prod(shape, dtype=np.uint64).item()
 
             if strides is None:
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -612,12 +612,9 @@ class Array:
 
                 shape = (shape,)
 
-            size = 1
-            for dim in shape:
-                size *= dim
-                if dim < 0:
-                    raise ValueError(
-                        f"negative dimensions are not allowed: {shape}")
+            size = np.prod(shape)
+            if size < 0:
+                raise ValueError(f"negative dimensions are not allowed: {shape}")
 
             if isinstance(size, np.integer):
                 size = size.item()
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -614,6 +614,8 @@ class Array:
 
             size = np.prod(shape)
             if size < 0:
+            # Previously the size was computed as "size = 1; size *= dim for dim in shape
+            # However it can fail when using data types, eg., numpy.uint64(1) * 2  returns 2.0 !
                 raise ValueError(f"negative dimensions are not allowed: {shape}")
 
             if isinstance(size, np.integer):
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -612,17 +612,18 @@ class Array:
 
                 shape = (shape,)
 
-            # Previously the size was computed as
+            shape_array = np.array(shape)
+
+            # Previously, the size was computed as
             #   "size = 1; size *= dim for dim in shape"
-            # However it can fail when using data types,
-            # eg numpy.uint64(1) * 2  returns 2.0 !
-            if np.any(np.array(shape) < 0):
+            # However this can fail when using certain data types,
+            # eg numpy.uint64(1) * 2 returns 2.0 !
+            if np.any(shape_array < 0):
                 raise ValueError(f"negative dimensions are not allowed: {shape}")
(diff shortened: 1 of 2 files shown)


pyopencl/array.py
```diff
@@ -612,12 +612,9 @@ def __init__(
...
-            for dim in shape:
-                size *= dim
-                if dim < 0:
-                    raise ValueError(
-                        f"negative dimensions are not allowed: {shape}")
+            size = np.prod(shape)
+            if size < 0:
+                raise ValueError(f"negative dimensions are not allowed: {shape}")
 
             if isinstance(size, np.integer):
```
inducer: With the change to `np.prod`, I would imagine this condition is always true.

pyopencl/array.py
```diff
@@ -612,12 +612,9 @@ def __init__(
...
 
-            size = 1
-            for dim in shape:
-                size *= dim
-                if dim < 0:
-                    raise ValueError(
-                        f"negative dimensions are not allowed: {shape}")
+            size = np.prod(shape)
+            if size < 0:
+                raise ValueError(f"negative dimensions are not allowed: {shape}")
```
inducer: Technically, this error will now not be triggered if there is an even number of negative shape axes.

pyopencl/array.py
```diff
@@ -612,12 +612,9 @@ def __init__(
 
                 shape = (shape,)
 
-            size = 1
-            for dim in shape:
-                size *= dim
-                if dim < 0:
-                    raise ValueError(
-                        f"negative dimensions are not allowed: {shape}")
+            size = np.prod(shape)
```
inducer: Could you leave a comment here about `uint64 * int`? I would not have suspected this, and I could easily see other folks working with this code falling into this trap.

inducer (commented):
Thanks! A few small wrinkles, but they should all be fixable with modest effort.

pyopencl/array.py
alexfikl: This should also handle the empty case, since
```
>>> np.prod((), initial=1)
1.0
```
always gives a float back?

pyopencl/array.py
alexfikl: We should probably add another check like `numpy`?
```
>>> np.empty((2.0, 5))
TypeError: 'float' object cannot be interpreted as an integer
```

alexfikl (commented):
Added a couple more comments. Nice catch on this!

pierrepaleo:
Thanks, a comment was added and the individual components are checked.

`numpy.prod(shape)` can still yield floating-point values, for example when `shape = (1, np.uint64(2), 4)`. 
So the product is computed on values cast to uint64, but not sure that's a good idea. Perhaps each component should be checked to be an int (and raise an error if a float is found). 

Any thoughts ?

pyopencl/array.py
```diff
@@ -612,15 +612,11 @@ def __init__(
...
-                    raise ValueError(
-                        f"negative dimensions are not allowed: {shape}")
-
-            if isinstance(size, np.integer):
-                size = size.item()
+            # Previously the size was computed as "size = 1; size *= dim for dim in shape
+            # However it can fail when using data types, eg., numpy.uint64(1) * 2  returns 2.0 !
+            if np.any(np.array(shape) < 0):
+                raise ValueError(f"negative dimensions are not allowed: {shape}")
+            size = np.prod(np.uint64(shape)).item()
```
inducer: Given the number of corner cases we discovered, I think it would be good to add some tests, so that we don't accidentally re-break things in the future.

inducer:
`numpy.prod` lets you specify the output dtype:

https://numpy.org/doc/stable/reference/generated/numpy.prod.html#numpy.prod

inducer:
Pushed some minor fixes. This looks good to me now. Thanks for your contribution!

PR #669, opened 2022-12-29
missing installation instructions for NVIDIA's GPUs

ghost:


doc/misc.rst
```diff
@@ -53,6 +53,10 @@ to install a CPU-based OpenCL driver.
 On macOS, pocl can offer a marked robustness (and, sometimes, performance)
 improvement over the OpenCL drivers built into the operating system.
 
+On Linux, to run on NVIDIA's GPUs::
+
+    conda install opencl-nvidia
```
inducer, line 58: Does this work? My understanding is that the ICD is tightly coupled to the driver version, and the driver is mostly in-kernel, i.e. unlikely to be installed by conda.

ghost: I installed latest Manjaro 22 with driver and CUDA installed automatically, but pyopencl wasn't working and this command (sudo pacman -S opencl-nvidia) enabled it. I don't think this is driver (nvidia driver 525.60.11), because that was installed ahead of time (nvidia driver + cuda). For more details, see:
https://www.randomguy.info/2022/12/ho-ho-ho-manjaro-22-is-here.html

So it did work for me. This command only enabled to run pyopencl on NVIDIA's GPU.

inducer: Does `opencl-nvidia` exist as a conda forge package? https://anaconda.org/conda-forge/opencl-nvidia does not seem to exist.

ghost: Hm. Yeah, I am not using conda unfortunately. But you guys don't have instructions for usage of pyopencl without conda and it took me quite a bit of search to find what I have to do...

inducer: Typically, the kernel driver comes with everything that's needed for OpenCL. All you may need to do is enable the systemwide ICDs in the conda environment, via

```
conda install ocl-icd-system
```

ghost: Interesting. Yeah, as I said I am not using conda... But when I try 'sudo pacman -S ocl-icd-system'
I get:

error: target not found: ocl-icd-system

So ocl-icd-system did not really provide me answer... Seems like conda is bit different.

inducer: Well, yes: Package names are not interchangeable between the arch package system, conda forge, and "regular" conda.

ghost: k. still, do you think this (sudo pacman -S opencl-nvidia) could be mentioned somewhere in the instructions so people know what to do in case they don't use conda and are on arch based linux to enable GPU usage?

inducer: Thanks for the suggestion, but I don't think it's feasible. There are too many linux distributions for us to be able to maintain install instructions on an ongoing basis.

There is some description of how things fit together at https://documen.tician.de/pyopencl/misc.html, plus instructions that should work everywhere, and I think that's about all we can reasonably provide.

ghost: That is correct (many distros) but in entire document you do not mention 'opencl-nvidia' package at all and it was the only thing that made it work for my environment. But ok. People have to search and suffer a bit I guess. By the way I tried your 'conda' instructions and that does not work at all on arch...

conda install pyopencl

gets me:

source ./miniconda3/bin/activate root
conda install pyopencl

Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.

**PackagesNotFoundError: The following packages are not available from current channels:**

  **- pyopencl**

Current channels:

  - https://repo.anaconda.com/pkgs/main/linux-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/linux-64
  - https://repo.anaconda.com/pkgs/r/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.

inducer: It appears you've installed `conda`/miniconda from anaconda, not from conda forge (as specified by the instructions). The two are not interchangeable.

inducer: I'll go ahead and add a comment to this effect.

ghost: Oh - ok. My bad. I thought I can install any. Uph. Sorry!

inducer: 99b619bf35691ac3db4a1f9805437067fa1c00c4

PR #682, opened 2023-04-25
docs: update intersphinx mapping from deprecated variant

alexfikl:
This was deprecated in Sphinx 6.2 and is now throwing some verbose warnings
https://www.sphinx-doc.org/en/master/changes.html#id7


inducer:
Thx!

Issue #681, opened 2023-04-14
Binary operations with Arrays of different memory layout


geggo:
For pyopencl.array.Array I get wrong results for binary operations such as adding if the memory layout differs for both arguments. Say, "A" is a C-contiguous, and "AF" is F-contiguous, then adding A+AF gives unexpected results. See this notebook for an example (tested with PyOpenCL 2022.3.1)

```python
import numpy as np
import pyopencl as cl
ctx = cl.create_some_context()
queue = cl.CommandQueue(ctx)
import pyopencl.array as cla
```
```python
# create simple numpy array 'a'
a = np.zeros((2,2))
a[0,1] = 1
a
```

    array([[0., 1.],
           [0., 0.]])

```python
# with different memory layout
af = np.asarray(a, order='F')
af
```

    array([[0., 1.],
           [0., 0.]])

```python
a + af
```

    array([[0., 2.],
           [0., 0.]])

```python
A = cla.to_device(queue, a)
AF = cla.to_device(queue, af)
```

```python
A, AF
```

    (cl.Array([[0., 1.],
            [0., 0.]]),
     cl.Array([[0., 1.],
            [0., 0.]]))

```python
# bug: A + AF gives different result than a + aF
B = A + AF
B
```

    cl.Array([[0., 1.],
           [1., 0.]])

Is this intended to work? Looking quickly at the source code, I could not see that the strides are taken into account. It seems that the underlying arrays are added in memory order. 

Thanks
Gregor

inducer:
Yep, this is a trade-off that I made at the outset, for two reasons:
- First, figuring out an appropriate launch config for strided accesses would add considerable overhead to an already too-hot code path leading up to arithmetic. (cf. numarray)
- Second, the existing code generation infrastructure won't easily stretch to true n-dimensional arrays. (https://github.com/inducer/loopy was intended, among other things, to address that use case)

It's obviously embarrassing that adding different-stride arrays silently gives wrong results, and I've been meaning to at least add a (C-based, to avoid a big impact on said hot code path) check to raise an error. PRs are definitely welcome along those lines.

That said, even if this issue were fixed, eagerly evaluating array calculations is still not a winning move, so I'm not super likely to devote much more attention than adding that check, cf. https://github.com/inducer/pytato/ for a possible remedy.

geggo:
[...]

PR #687, opened 2023-05-17
elwise: support more shapes for binary ops

matthiasdiener:
Looking for feedback - I'm not sure this approach is viable (e.g., due to potential issues with contiguity).


inducer:
Pyopencl arrays don't do proper broadcasting, see #681.

PR #702, opened 2023-10-06
Run flake8 without an env on Github CI

inducer:
In a way, this addresses
https://github.com/zheller/flake8-quotes/issues/117, since it forces the Python that runs flake8 onto 'oldest supported', currently 3.8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,6 @@ jobs:
             run: |
                 curl -L -O https://tiker.net/ci-support-v0
                 . ci-support-v0
-                build_py_project_in_conda_env
                 install_and_run_flake8 "$(get_proj_name)" examples/*.py test/*.py
 
     pylint:


alexfikl:
Why was this picking up python 3.12? The CI already sets `python-version: '3.8'` in the `flake8` job for compatibility..

inducer:
That 3.8 specified the Python version installed by Github, which was shortly thereafter made moot by the conda env being installed.

alexfikl:
Hmm, that's a bit confusing.. Should we do a `conda env create python=$PYTHON_VERSION` or something like that in `ci-support`?

Only `pytools` seems to be doing a matrix on the `python-version` and it uses `venv` directly, so that works. Most other projects try to run `flake8` on `python-version: '3.8'` with conda and probably have the same issue as this, but it didn't complain yet..

inducer:
IDK. Kind of reluctant to attempt to overengineer something here.

PR #704, opened 2023-10-12
Enable Py3.12 in cibuildwheel

inducer:


inducer:
https://github.com/inducer/islpy/releases/tag/v2023.2.5, should be on the package index sometime soon.

PR #713, opened 2024-02-02
only warn about argument sizes on Nvidia GPUs

matthiasdiener:
In our experience, only Nvidia GPUs actually have compilation errors when the argument size is too large; most other CL devices have artificially low limits (usually 1024 bytes, roughly corresponding to the 127 arguments requirement of the [C standard](https://c0x.shape-of-code.com/5.2.4.1.html)), but compile and run much larger argument sizes fine, thus leading to lots of spurious warnings on any non-Nvidia device.

_Please squash_

### DIFF
--- a/pyopencl/invoker.py
+++ b/pyopencl/invoker.py
@@ -319,7 +319,7 @@ def _get_max_parameter_size(dev):
     dev_limit = dev.max_parameter_size
     pocl_version = get_pocl_version(dev.platform, fallback_value=(1, 8))
     if pocl_version is not None and pocl_version < (3, 0):
-        # Current PoCL versions (as of 04/2022) have an incorrect parameter
+        # Older PoCL versions (<3.0) have an incorrect parameter
         # size limit of 1024; see e.g. https://github.com/pocl/pocl/pull/1046
         if dev_limit == 1024:
             if dev.type & cl.device_type.CPU:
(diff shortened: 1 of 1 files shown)


pyopencl/invoker.py
inducer, line 1: Could you explain the context a bit more? Were there spurious warnings that were bothersome?

pyopencl/invoker.py
```diff
@@ -332,8 +332,14 @@ def _check_arg_size(function_name, num_cl_args, arg_types, devs):
     """Check whether argument sizes exceed the OpenCL device limit."""
 
     for dev in devs:
-        dev_ptr_size = int(dev.address_bits / 8)
         dev_limit = _get_max_parameter_size(dev)
+        if dev_limit != 4352:
```
inducer: I'm not sure I like this check, it seems too broad. I expect that AMD and Intel GPUs would have some limit as well, but these changes would avoid warnings in those cases, even if they're applicable. I would be more willing to limit this to GPU devices, which is testable via flags in the device info.

pyopencl/invoker.py
matthiasdiener: In our experience, only Nvidia GPUs actually have compilation errors when the argument size is too large; most other CL devices have artificially low limits (usually 1024 bytes, roughly corresponding to the 127 arguments requirement of the [C standard](https://c0x.shape-of-code.com/5.2.4.1.html)), but compile and run much larger argument sizes fine, thus leading to lots of spurious warnings on any non-Nvidia device.

pyopencl/invoker.py
matthiasdiener: I changed the check in [eeeb8e8](https://github.com/inducer/pyopencl/pull/713/commits/eeeb8e8a09976704c84c6652adef4aeef9bdad79), what do you think?

PR #716, opened 2024-03-19
Compiler cache: honor XDG_CACHE_HOME on MacOS

matthiasdiener:
Similar to https://github.com/inducer/pytools/pull/204

### DIFF
--- a/pyopencl/cache.py
+++ b/pyopencl/cache.py
@@ -348,20 +348,20 @@ def _create_built_program_from_source_cached(ctx, src, options_bytes,
         except ImportError:
             import appdirs
 
-        cache_dir = join(appdirs.user_cache_dir("pyopencl", "pyopencl"),
+        # Determine the cache directory in the same way as pytools.PersistentDict,
+        # which PyOpenCL uses for invoker caches.
+        if sys.platform == "darwin" and os.getenv("XDG_CACHE_HOME") is not None:
+            # appdirs and platformdirs do not handle XDG_CACHE_HOME on macOS
+            # https://github.com/platformdirs/platformdirs/issues/269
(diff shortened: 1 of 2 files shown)


alexfikl:
I understand the need here, but I'm a curious about some things:
* Why only on macOS and not Windows? Can the Windows paths be overwritten by environment variables?
* If we're forcing a non-native environment variable on other OSes, why not use something like `PYOPENCL_CACHE_DIR` (to mirror `PYOPENCL_NO_CACHE`)?

matthiasdiener:
> * Why only on macOS and not Windows? Can the Windows paths be overwritten by environment variables?

Looking at https://github.com/platformdirs/platformdirs/blob/main/src/platformdirs/windows.py#L251, I guess so, but my knowledge of Windows is almost zero 😞 

> * If we're forcing a non-native environment variable on other OSes, why not use something like `PYOPENCL_CACHE_DIR` (to mirror `PYOPENCL_NO_CACHE`)?

We could add `PYOPENCL_CACHE_DIR`, but the current goal was to make the disk caching selection consistent across loopy and pyopencl, for both Mac and Linux. 

Edit: another important reason: _Pocl_ also honors `XDG_CACHE_HOME` on MacOS.

alexfikl:
> Edit: another important reason: _Pocl_ also honors `XDG_CACHE_HOME` on MacOS.

Fair enough, but they just fallback to it and also have a [`POCL_CACHE_DIR`](http://portablecl.org/docs/html/env_variables.html) (at least that's what the docs say).

matthiasdiener:
This is ready for review @inducer.

inducer:
LGTM, thanks!

PR #718, opened 2024-03-28
Port to np.random.Generator

alexfikl:
This ports the examples to `np.random.Generator` to fix some deprecation warnings in numpy 2.0.

I also updated `gl_particle_animation` to not do `from GL import *` because it was giving a lot of pylint warnings. I couldn't run it though because it raised
```
AttributeError: type object 'pyopencl._cl.context_properties' has no attribute 'GL_CONTEXT_KHR'
``` 
Not quite sure what's missing on my system (I compiled with `CL_ENABLE_GL` and whatever else I could find).

### DIFF
--- a/examples/demo.py
+++ b/examples/demo.py
@@ -3,8 +3,9 @@
 import numpy as np
 import pyopencl as cl
 
-a_np = np.random.rand(50000).astype(np.float32)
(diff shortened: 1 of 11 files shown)


inducer:
Thanks!

PR #723, opened 2024-04-02
Port deprecated pkg_resources to importlib.resources

alexfikl:
This ports from `pkg_resources` to the newer `importlib.resources` as described in
https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename
and removes an implicit runtime dependency on `setuptools`. It also needs a newer version of `importlib.resources` because the pre 3.9 version didn't support directories and `pyopencl/cl` is a directory in our case.

Not sure how to test this because the include path is already in the first path it looks at: `join(abspath(dirname(__file__)), "cl")`. I checked that the `pkg_resources` result is the same on my system. Does the CI test that at all?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,8 @@ jobs:
                 . ci-support-v0
                 ./configure.py --cl-use-shipped-ext
                 build_py_project_in_conda_env
-                python -m pip install mypy types-setuptools
+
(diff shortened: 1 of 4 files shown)


inducer:
I think this was added for some bizarre systems where the two disagree, I don't remember the specifics.

> Does the CI test that at all?

No, not AFAIK.

Thanks for doing this, this is good enough for me.

PR #732, opened 2024-04-19
fix create_some_context doc

matthiasdiener:


matthiasdiener:
Side question: Is the comment 

https://github.com/inducer/pyopencl/pull/732/files#diff-c92a2dfd89b4e37e39a4e9664d9a0f65e5a9a332c741e429c225b69f8bc15d7aR125-L129

> Calling the constructor with no arguments will fail for recent
>         CL drivers that support the OpenCL ICD. If you want similar,
>         just-give-me-a-context-already behavior, we recommend
>         :func:`create_some_context`. See, e.g. this
>         `explanation by AMD       <https://web.archive.org/web/20101114195033/https://developer.amd.com/support/KnowledgeBase/Lists/KnowledgeBase/DispForm.aspx?ID=71>`__.

still correct?

doc/runtime_platform.rst
```diff
@@ -179,7 +179,7 @@ Context
 
     |comparable|
 
-.. function:: create_some_context(interactive=True, answers=None, cache_dir=None)
+.. function:: create_some_context(interactive=True, answers=None)
```
inducer: Make this an `autofunction` instead.

inducer:
> Side question: Is the comment still correct?

Yes, but it's very fundamentally outdated, as no-one uses OpenCL without ICD loaders any more.

matthiasdiener:
> > Side question: Is the comment still correct?
> 
> Yes, but it's very fundamentally outdated, as no-one uses OpenCL without ICD loaders any more.

I reworded this slightly in 43280c2 (but I'm not sure if it's correct).

doc/runtime_platform.rst
matthiasdiener: Done in [cd38c09](https://github.com/inducer/pyopencl/pull/732/commits/cd38c091a36830656004e18ae494c8e6a77fe157)

doc/runtime_platform.rst
```diff
@@ -179,12 +179,12 @@ Context
 
     |comparable|
 
-.. function:: create_some_context(interactive=True, answers=None)
+.. autofunction:: create_some_context
 
     Create a :class:`Context` 'somehow'.
```
inducer: Please also move this to the function docstring. :)

matthiasdiener: :) What do you think of [9beccee](https://github.com/inducer/pyopencl/pull/732/commits/9beccee24e326438587c95ef6347966244866054)?

doc/runtime_platform.rst
```diff
@@ -122,9 +122,9 @@ Context
 
     .. note::
 
-        Calling the constructor with no arguments will fail for recent
-        CL drivers that support the OpenCL ICD. If you want similar,
-        just-give-me-a-context-already behavior, we recommend
+        Calling the constructor with no arguments will fail for
+        CL drivers that support the OpenCL ICD (which applies to most modern drivers).
```
inducer:
```suggestion
        CL drivers that support the OpenCL ICD (which applies to most modern systems).
```

inducer:
Thx!

PR #735, opened 2024-04-19
add choose_devices()

matthiasdiener:


pyopencl/__init__.py
```diff
@@ -1576,7 +1577,7 @@ def create_some_context(interactive: Optional[bool] = None,
             from pyopencl.tools import get_test_platforms_and_devices
             for _plat, devs in get_test_platforms_and_devices():
                 for dev in devs:
-                    return Context([dev])
```
matthiasdiener, line 1579: This code looks a bit strange, why not just return `Context([devs[0]])`?

alexfikl: That wouldn't work if `devs` is empty?

matthiasdiener: Oh, I see. Is that a Python pattern? I would just have written `if devs: return ...`

alexfikl: I don't think it's a common pattern, but it avoids an `if` in Python. Not sure if that's much faster + this isn't a hot path..

pyopencl/__init__.py
```diff
@@ -1552,10 +1552,10 @@ def __init__(self, ctx, size, alignment, flags, *, _interface,
 
 # {{{ create_some_context
 
-def create_some_context(interactive: Optional[bool] = None,
-                        answers: Optional[List[str]] = None) -> Context:
+def choose_device(interactive: Optional[bool] = None,
+                  answers: Optional[List[str]] = None) -> List[Device]:
```
alexfikl: Naming wise, it's weird that this returns a list and it's called `choose_device`?

matthiasdiener: Yeah, we could rename it to `choose_devices` or so. My only concern was that most (?) people will just choose 1 device...

Issue #731, opened 2024-04-12
`build program` times increasing with rank count on Mac when caching is enabled


majosm:
The times reported by the `build program: kernel '<name>' was part of a lengthy source build resulting from a binary cache miss (<time>)` output appear to increase fairly dramatically with rank count on my Mac with caching enabled, even when using rank-local cache directories. For example, when running the `wave-op-mpi` example in grudge, with 16 ranks and caching disabled via `PYOPENCL_NO_CACHE=1`, I see:

```
INFO:pyopencl:build program: kernel 'frozen_nodes0_2d' was part of a lengthy uncached source build (cache disabled by user) (0.31 s)
INFO:pyopencl:build program: kernel 'frozen_nodes0_2d' was part of a lengthy uncached source build (cache disabled by user) (0.31 s)
INFO:pyopencl:build program: kernel 'frozen_nodes0_2d' was part of a lengthy uncached source build (cache disabled by user) (0.32 s)
INFO:pyopencl:build program: kernel 'frozen_nodes0_2d' was part of a lengthy uncached source build (cache disabled by user) (0.32 s)
```
With caching enabled (and empty cache) I see:

```
INFO:pyopencl:build program: kernel 'rhs' was part of a lengthy source build resulting from a binary cache miss (0.49 s)
INFO:pyopencl:build program: kernel 'rhs' was part of a lengthy source build resulting from a binary cache miss (0.68 s)
INFO:pyopencl:build program: kernel 'rhs' was part of a lengthy source build resulting from a binary cache miss (2.94 s)
INFO:pyopencl:build program: kernel 'rhs' was part of a lengthy source build resulting from a binary cache miss (3.28 s)
```
(Note: `rhs` is missing from the first output, presumably because the time is below the output threshold. The lack of `frozen_nodes0_2d` in the second output is confusing though.)

If I increase to 16 ranks, with no caching I see:

```
INFO:pyopencl:build program: kernel 'frozen_nodes0_2d' was part of a lengthy uncached source build (cache disabled by user) (0.52 s)
INFO:pyopencl:build program: kernel 'frozen_nodes0_2d' was part of a lengthy uncached source build (cache disabled by user) (0.54 s)
INFO:pyopencl:build program: kernel 'frozen_nodes0_2d' was part of a lengthy uncached source build (cache disabled by user) (0.56 s)
[...]

PR #736, opened 2024-04-22
invoker, scan: honor PYOPENCL_NO_CACHE for writing

matthiasdiener:
(found while looking at #731)

With this PR, no Pyopencl cache is created when `PYOPENCL_NO_CACHE` is set.

### DIFF
--- a/pyopencl/invoker.py
+++ b/pyopencl/invoker.py
@@ -373,7 +373,8 @@ def _check_arg_size(function_name, num_cl_args, arg_types, devs):
 # }}}
 
 
-invoker_cache = WriteOncePersistentDict(
+if not cl._PYOPENCL_NO_CACHE:
+    invoker_cache = WriteOncePersistentDict(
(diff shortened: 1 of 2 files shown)


pyopencl/invoker.py
```diff
@@ -373,7 +373,8 @@ def _check_arg_size(function_name, num_cl_args, arg_types, devs):
 # }}}
 
 
-invoker_cache = WriteOncePersistentDict(
+if not cl._PYOPENCL_NO_CACHE:
+    invoker_cache = WriteOncePersistentDict(
         "pyopencl-invoker-cache-v41",
         key_builder=_NumpyTypesKeyBuilder(),
         in_mem_cache_size=0)
```
matthiasdiener, line 380: out of curiosity, why isn't this using the in-mem cache?

Edit: I think the invoker cache entry will be read only once, so additional in-mem caching might not be useful.

PR #742, opened 2024-05-01
Allow, test zero-sized SVM allocations

inducer:


inducer:
Merging despite Mac failure, see #743.

Issue #743, opened 2024-05-01
Pocl Mac crashing again


inducer:
See e.g. [this CI log from main on April 27](https://github.com/inducer/pyopencl/actions/runs/8864701008/job/24340155005#step:3:1827).

This was OK as recently as [April 25](https://github.com/inducer/pyopencl/actions/runs/8838555600/job/24269968782).

@matthiasdiener, any idea what's up here?

matthiasdiener:
I don't know why it is crashing, but OS and arch have changed between those runs:

- April 25: macOS 12.7.4, x86_64
- April 27: macOS 14.4.1, ARM

(GitHub seems to be migrating mac CI to ARM)

matthiasdiener:
Seems like the ARM runners have less memory than x86_64, see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

inducer:
Do the tests work locally on your M1?

matthiasdiener:
> Do the tests work locally on your M1?

Yes, everything works fine on my M1 with pocl 5 from conda-forge (with or without xdist).

PR #744, opened 2024-05-01
workaround Mac CI

matthiasdiener:
Works around #743

_Please squash_


inducer:
Thanks for figuring out that this is Ranlux's fault. Even better: #745. Closing here.

PR #746, opened 2024-05-02
readme: make badges horizontal again

alexfikl:
Stole this from setuptools. Not sure if Github is going to fix it any time soon..

For reference, here's the README rendered:
https://github.com/alexfikl/pyopencl/blob/fix-readme/README.rst


inducer:
Huh, thanks! Though, I'll be honest, the vertical badges don't bother me much.

PR #738, opened 2024-04-23
Disable build caching when using pocl

majosm:
Fixes #731.

PR #749, opened 2024-05-07
Disable source build cache on pocl/nvidia

matthiasdiener:
Followup of #738.

_Please squash_

### DIFF
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -1550,10 +1550,6 @@ def choose_devices(interactive: Optional[bool] = None,
     :returns: a list of :class:`Device` instances.
     """
 
-    # This must be done before PoCL is actually loaded
-    from pyopencl.characterize import setup_pocl_cache_dirs
-    setup_pocl_cache_dirs()
-
     if answers is None:
         if "PYOPENCL_CTX" in os.environ:
             ctx_spec = os.environ["PYOPENCL_CTX"]
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -493,8 +493,16 @@ class Program:
             cache_dir = getattr(self._context, "cache_dir", None)
 
         build_descr = None
-        if _PYOPENCL_NO_CACHE and self._prg is None:
-            build_descr = "uncached source build (cache disabled by user)"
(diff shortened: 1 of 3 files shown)


matthiasdiener:
~~[1a9ad0d](https://github.com/inducer/pyopencl/pull/749/commits/1a9ad0dc89f1d66df750f71374fcb14604eec0cd) outlines a way to ensure pocl uses PID-local caches.~~ This is ready for a first look @inducer @majosm

pyopencl/__init__.py
```diff
@@ -1540,6 +1548,10 @@ def choose_devices(interactive: Optional[bool] = None,
     :returns: a list of :class:`Device` instances.
     """
 
+    # This must be done before PoCL is actually loaded
+    from pyopencl.characterize import setup_pocl_cache_dirs
+    setup_pocl_cache_dirs()
```
inducer: Could you explain why this is here? This seems to defeat inter-run caching, which, in my book, is definitely desirable.

matthiasdiener: This is a workaround for the intermittent failures we have seen as part of #738. 

I couldn't think of a way to ensure PID-local caches that also support inter-run caching, without resorting to using MPI rank numbers (which wouldn't even help for the CI errors seen in #738).

Edit: some sources regarding caching issues in pocl: https://github.com/pocl/pocl/issues/430, https://github.com/pocl/pocl/pull/466

inducer: Before we go in this direction, I think we should engage with the pocl folks and least file an issue to see what they think about fixing concurrent access to their caches.

matthiasdiener: Reported here: https://github.com/pocl/pocl/issues/1480, including a small reproducer.

matthiasdiener: Reverted in [b368eba](https://github.com/inducer/pyopencl/pull/749/commits/b368eba29f942b3e1939489f131a3799bbfbe37e)

matthiasdiener:
This is ready for review @inducer. I have not noticed any cache concurrency issues with this PR+the new pocl version (`5.0 * _5`).

PR #755, opened 2024-05-10
use scikit-build-core

### LINKED ISSUE #753: `pip install .` causes messy directory layout
**Describe the bug**
`pip install .` seems to write pyopencl files into the main environment:

```
$ pip install .
[...]
$ pip uninstall pyopencl
Found existing installation: pyopencl 2024.2.2
Uninstalling pyopencl-2024.2.2:
  Would remove:
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/CITATION.cff
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/CMakeLists.txt
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/LICENSE
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/Makefile.in
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/README.rst
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/README_SETUP.txt
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/aksetup_helper.py
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/configure.py
    /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/contrib/cldis.py
[...]
```

**To Reproduce**
1. `$ cd pyopencl`
2. `$ pip install .`
3. `$ pip uninstall pyopencl`

**Expected behavior**
Keep the files in the `pkgs/pyopencl-*` subdirectory.

 - OS: MacOS 
  - PyOpenCL version: 2024.2.2

matthiasdiener:
This appears to fix #753.

TODO:

- [x] Nicer config summary 
- [x] Nicer way to configure?
- [x] MacOS links against Mac CL framework
- [x] Windows
- [x] pylint (#757)
- [x] update docs
- [x] undo CI changes
- [x] Pass CI on Gitlab: [Branch on Gitlab](https://gitlab.tiker.net/inducer/pyopencl/-/commits/scikit-build-core)

_Please squash_


alexfikl:
If we want to switch, why not `meson-python` instead? The numpy / scipy / matplotlib ecosystem seems to have gone that way and it's probably a good idea to follow..

matthiasdiener:
> If we want to switch, why not `meson-python` instead? The numpy / scipy / matplotlib ecosystem seems to have gone that way and it's probably a good idea to follow..

TIL, thanks! I think the main reason for using `scikit-build-core` is that it is what nanobind recommends, and it's used in their examples.

matthiasdiener:
This seems to work fine and is ready for a first look @inducer. I'm not sure what the best way to replace/reimplement the `configure.py` functionality is.

alexfikl:
> This seems to work fine and is ready for a first look @inducer. I'm not sure what the best way to replace/reimplement the `configure.py` functionality is.

What functionality exactly? Generally the configuration can be done in `CMakeLists.txt` and you can pass it arguments using
```
pip install . --config-settings='cmake.args=-DSOME_DEFINE=ON;-DOTHER=OFF'
```

pyproject.toml
```diff
@@ -1,16 +1,65 @@
...
+]
+build-backend = "scikit_build_core.build"
+
+[tool.scikit-build.cmake.define]
+PYOPENCL_TRACE = {env="CL_TRACE", default="OFF"}
+PYOPENCL_ENABLE_GL = {env="CL_ENABLE_GL", default="OFF"}
+PYOPENCL_USE_SHIPPED_EXT = {env="CL_USE_SHIPPED_EXT", default="OFF"}
+OpenCL_INCLUDE_DIR = {env="CL_INC_DIR"}
+PYOPENCL_CL_LIB_DIRS = {env="CL_LIB_DIR"}
+OpenCL_LIBRARY = {env="CL_LIBNAME"}
```
alexfikl: These should also be defined in `CMakeLists.txt` as `option(PYOPENCL_TRACE "Enable tracing functionality" OFF)`? At least that's idiomatic from what I recall..

.github/workflows/ci.yml
```diff
@@ -33,14 +33,22 @@ jobs:
...
 
                 curl -L -O https://tiker.net/ci-support-v0
                 . ci-support-v0
-                build_py_project_in_conda_env
+
+                install_conda_deps
+
+                export CL_INC_DIR="$CONDA_PREFIX/include"
+                export CL_LIB_DIR="$CONDA_PREFIX/lib"
+                export CL_LIBNAME=OpenCL
```
alexfikl: Could this logic be in `CMakeLists.txt`? Something like `if($ENV{CONDA_PREFIX}) set(CL_INC_DIR "${CONDA_PREFIX}/include")`?

matthiasdiener:
> What functionality exactly? Generally the configuration can be done in `CMakeLists.txt` and you can pass it arguments using

More specifically, I meant an easy-ish way for the user to configure the build, similar to the `configure.py` / `siteconf.py` system in the current version.


> ```
> pip install . --config-settings='cmake.args=-DSOME_DEFINE=ON;-DOTHER=OFF'
> ```

Yeah, that is an option for replacing `configure.py`. Additionally, I was thinking of providing a `siteconf.cmake` file similar to `siteconf.py`.

.github/workflows/ci.yml
matthiasdiener: I'm not sure, it feels a bit strange for the `CMakeLists.txt` file to know about conda.

inducer: I agree, but maybe the logic could be made conditional on there being an OpenCL library in the conda env? If there is, I think there is an overwhelming likelihood it should be used. And PyOpenCL's documentation talks about conda as a means to install, so it's not like the two projects are totally unrelated.

inducer:
I think ditching the `configure.py` mechanism is the right approach at this point. As long as we document the `cmake.args` way (also maybe add an entry to the changelog), I think that's OK.

.github/workflows/ci.yml
alexfikl: For context, on my system (cmake 3.29.3) there are a few cmake modules that look at `CONDA_PREFIX` in particular: `GNUInstallDirs.cmake`, `UnixPaths.cmake` and `FindPython/Support.cmake`.

So I don't think there's anything strange about teaching the build a bit more about the platform, which in this case is `conda`.

alexfikl: Also, looking at `FindOpenCL.cmake`, it seems like we can set `OCL_ROOT=$CONDA_PREFIX` to let it find the right folders itself.

matthiasdiener:
Here are the OpenCL config options that we have been discussing (all except nr. 4 should work with the current state of this PR):

1. Get path from `CONDA_PREFIX` if set
2.  ~~`siteconf.cmake`~~
3. Environment variables
4. Having some kind of `configure.py`-like frontend (that calls pip install with the right options?)
5. `pip install . --config-settings='cmake.args=-D....'`
6. Fallback: let's just have cmake try to find OpenCL

If I understood correctly, option 4 isn't really desired. Are there any other options we should add/remove? (e.g., perhaps `sieconf.cmake` isn't necessary?)


Besides the question above, this is ready for a second look (in particular, regarding the next steps).

.github/workflows/ci.yml
matthiasdiener: TIL, thanks. conda support was added in [9b3c600](https://github.com/inducer/pyopencl/pull/755/commits/9b3c6006492b298882bcb0630389382e5262cd53)

pyproject.toml
matthiasdiener: Done in [34a47d9](https://github.com/inducer/pyopencl/pull/755/commits/34a47d97e355791c8dfbf99174a0da168124482a)

matthiasdiener:
This is ready for review @inducer/@alexfikl.

.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
             run: |
                 curl -L -O https://tiker.net/ci-support-v0
                 . ci-support-v0
-                ./configure.py --cl-use-shipped-ext
+                export CL_USE_SHIPPED_EXT=on
```
alexfikl, line 60: Mostly curious @inducer, why aren't the extensions enabled by default? Not very portable?

.github/workflows/ci.yml
```diff
@@ -110,8 +110,15 @@ jobs:
...
                 curl -L -O https://tiker.net/ci-support-v0
                 . ci-support-v0
-                ./configure.py --cl-use-shipped-ext
-                build_py_project_in_conda_env
+
+                install_conda_deps
+
+                export CL_USE_SHIPPED_EXT=on
+                export CL_INC_DIR="$CONDA_PREFIX/Library/include"
+                export CL_LIB_DIR="$CONDA_PREFIX/Library/lib"
```
alexfikl: Why does this still require exporting the variables here? Does cmake not find them? Might be worth adding a note somewhere if this is actual issue.

CMakeLists.txt
```diff
@@ -7,19 +7,86 @@ if (NOT SKBUILD)
...
+             )
+  if(NOT ${git_result} EQUAL 0)
+    execute_process(COMMAND git describe --long --always --dirty=*
+             OUTPUT_VARIABLE PYOPENCL_VERSION_GIT
+             OUTPUT_STRIP_TRAILING_WHITESPACE
+             )
+    set(PYOPENCL_REL "(dev)")
+  else()
+    set(PYOPENCL_REL "(release)")
+  endif()
```
alexfikl, line 125: This `PYOPENCL_REL` seems to be just for display purposes? Is it really needed?

EDIT: The `cmake` output only shows up on `pip install --verbose` anyway (right?), so this should just be a nice debug variable, which sounds good!

CMakeLists.txt
```diff
@@ -7,19 +7,86 @@ if (NOT SKBUILD)
...
+# {{{ Options
+
+option(PYOPENCL_TRACE "Enable OpenCL tracing" OFF)
+option(PYOPENCL_ENABLE_GL "Enable OpenGL interoperability" OFF)
+option(PYOPENCL_USE_SHIPPED_EXT "Use shipped CL extension header" OFF)
+set(CL_INC_DIR CACHE STRING "OpenCL include directory")
+set(CL_LIB_DIR CACHE STRING "OpenCL library directory")
+set(CL_LIBNAME CACHE STRING "OpenCL library name")
+
+if(NOT CL_INC_DIR)
```
alexfikl: I'm not sure I follow all this logic here, since it seems to overwrite the variables a few times. What's the priority order? I would expect something like
1. Whatever is passed to `cmake -DCL_LIB_DIR=/somedir` (or equivalently through pip `--config-settings`)
2. If not given, look at some standard environment variable like `OpenCL_ROOT` / `OCL_ROOT`
3. If not given, look at the environment variable `CL_LIB_DIR`.
5. If still not given, try to look at conda.
6. Let `FindOpenCL` try to find it.
7. Error!!!

It seems like this first tries to set it from `conda` (line 24) and then overwrites it with the environment variable (line 32).

CMakeLists.txt
```diff
@@ -7,19 +7,86 @@ if (NOT SKBUILD)
...
+else()
+  set(PYOPENCL_VERSION_GIT "v${PROJECT_VERSION}")
 endif()
-find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
+
+# }}}
+
+# {{{ Constrain FindPython to find the Python version used by scikit-build
+
+message(STATUS "Python_VERSION ${PYTHON_VERSION_STRING}")
```
alexfikl: Who sets `PYTHON_VERSION_STRING` and these over variables? `scikit-build-core`?

CMakeLists.txt
```diff
@@ -7,19 +7,86 @@ if (NOT SKBUILD)
...
+    set(PYOPENCL_REL "(release)")
+  endif()
+else()
+  set(PYOPENCL_VERSION_GIT "v${PROJECT_VERSION}")
 endif()
-find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
+
+# }}}
+
+# {{{ Constrain FindPython to find the Python version used by scikit-build
```
alexfikl:
```suggestion
# {{{ Constrain FindPython to find the Python version used by scikit-build-core
```
:grin:

CMakeLists.txt
```diff
@@ -39,13 +112,15 @@ execute_process(
...
+include_directories(${CL_INC_DIR} ${Python_NumPy_INCLUDE_DIRS})
 
 find_package(nanobind CONFIG REQUIRED)
-find_package(NumPy REQUIRED)
-find_package(OpenCL REQUIRED)
 
-include_directories(${OpenCL_INCLUDE_DIR} ${NumPy_INCLUDE_DIRS})
+set(OpenCL_ROOT ${CL_LIB_DIR})
+set(OpenCL_INCLUDE_DIR ${CL_INC_DIR})
+set(OpenCL_LIBRARY ${CL_LIBNAME})
```
alexfikl, line 164: Doesn't `find_package(OpenCL)` set `OpenCL_LIBRARY` as an output?

pyproject.toml
```diff
@@ -1,16 +1,57 @@
...
+pocl = [
+    "pocl_binary_distribution>=1.2",
+]
+test = [
+    "Mako",
+    "pytest>=7.0.0",
+]
+
+[project.urls]
+Homepage = "https://mathema.tician.de/software/pyopencl"
```
alexfikl, line 54:
```suggestion
[project.urls]
Homepage = "https://mathema.tician.de/software/pyopencl"
Documentation = "https://documen.tician.de/pyopencl"
Repository = "https://github.com/inducer/pyopencl"
```
? I'm not quite sure how many of these `PyPI` actually uses..

alexfikl (commented):
Looks a lot nicer! :heart: 

I'm mostly still a bit confused about the order of those `CL_LIB_DIR` and friends variables, but otherwise it's just nitpicks.

CMakeLists.txt
matthiasdiener: Good question, as far as I can tell, this part of the code isn't functional (e.g., `PYTHON_VERSION_STRING` isn't set at all). Removed in 08add61.

CMakeLists.txt
matthiasdiener: Right, this is just for the config summary shown at the end (needs `--verbose`).

.github/workflows/ci.yml
matthiasdiener: The problem is, I don't know what the OpenCL library is called on Windows, so I had to revert the commit that added conda detection on Windows: https://github.com/inducer/pyopencl/pull/755/commits/e4906a63548e646c40e20b61df0f7319568ea606

matthiasdiener: found it :-) [96814a2](https://github.com/inducer/pyopencl/pull/755/commits/96814a222df3bec7ceee1d52d9c1199a8690f8ef)

CMakeLists.txt
matthiasdiener: I think what you describe roughly matches what the implementation is (minus 2.). I'll clarify the code.

pyproject.toml
matthiasdiener: I'm fine with this, but I think @inducer should decide.

CMakeLists.txt
matthiasdiener: The effect of setting these variables is that it constrains FindOpenCL below to the options selected by the user.

CMakeLists.txt
matthiasdiener: [4ebf7e9](https://github.com/inducer/pyopencl/pull/755/commits/4ebf7e9468d2a13c829e82fe2ce30717acd0bcc3) clarifies this hopefully. The order of precedence currently is:

1. Whatever is passed as a CMake option, e.g. via `pip --config-settings`
2. If not given, look at the environment variables `CL_INC_DIR`, `CL_LIB_DIR`, `CL_LIBNAME`
3. If not, try looking at `CONDA_PREFIX`
4. If not, let `FindOpenCL` find it in its default locations (this can work if OpenCL is installed system-wide, like on MacOS).
5. If not, error.

CMakeLists.txt
alexfikl: Ah, that's cool! I thought `find_library(OpenCL_LIBRARY ...)` would overwrite the variable, but apparently it just skips the search. Thanks!

CMakeLists.txt
alexfikl: This looks good to me now!

I only mentioned the `OCL_ROOT` variable because the version of `FindOpenCL` on my system (cmake `3.29.3`) has a `ENV OCL_ROOT` in the `find_library` hints, so I assumed that may be a more general variable.

alexfikl (approved):

CMakeLists.txt
```diff
@@ -1,35 +1,126 @@
-cmake_minimum_required(VERSION 3.17...3.22)
+cmake_minimum_required(VERSION 3.17...3.26)
```
alexfikl, line 1: Is the upper bound necessary here? It doesn't seem like cmake complains about this, so not quite sure what it's for..

CMakeLists.txt
```diff
@@ -1,35 +1,126 @@
...
+
+project(pyopencl LANGUAGES CXX)
```
alexfikl:
```suggestion
project(pyopencl LANGUAGES CXX VERSION ${SKBUILD_PROJECT_VERSION})
```
? I think `PROJECT_VERSION` is used below and it would be empty otherwise.

CMakeLists.txt
matthiasdiener: This was copied from the nanobind example. Like you said, I think this has no real effect in practice.

CMakeLists.txt
matthiasdiener: Great catch, thanks!

.github/workflows/ci.yml
inducer: Technically, each system is supposed to have an up-to-date header describing its available CL extensions. That could include extensions that aren't known upstream. Practically, those headers are often out of date and barely maintained. Still, the correct approach is to rely on the system header. Using a shipped header is at best a stopgap.

alexfikl: Ah, I see. Thank you for explaining!

inducer:
I've pushed the branch to Gitlab to check CI there (added a link to the description for easy access). While `main` currently doesn't pass on Gitlab either :cry:, there are some issues there that might merit a look, particularly the downstream failures.

matthiasdiener:
> I've pushed the branch to Gitlab to check CI there (added a link to the description for easy access). While `main` currently doesn't pass on Gitlab either 😢, there are some issues there that might merit a look, particularly the downstream failures.

I think the 'downstream' Gitlab tests should be fixed now ('examples'/'K40' tests fail with `main` on Gitlab too), but I don't see a way to retrigger the CI run on Gitlab.

matthiasdiener:
Not sure what's going on with the boxtree test (failure seems to be spurious?), but gitlab tests seem to be fixed.

inducer:
We can't really get a clean test run on Gitlab because of https://github.com/inducer/pytools/issues/227, so I'll put this on hold until we can.

inducer:
Nvm, we can, now that the release is yanked.

inducer:
LGTM. Thanks for making this happen. Let's do it!

PR #757, opened 2024-05-16
address pylint>=3.2 issues

matthiasdiener:


inducer:
Could you use error names rather than numbers?

matthiasdiener:
> Could you use error names rather than numbers?

Done in 013150a (but looks a bit clumsy).

inducer:
Pushed some minor tweaks, merging.

inducer:
Thanks!

PR #761, opened 2024-05-20
test_bitonic_sort: use ctx_factory()

matthiasdiener:
(otherwise, the wrong context may be used, and users get asked to select a context with `pytest -s`)


inducer:
Thx!

PR #768, opened 2024-06-08
set sync_safe=False for caches

matthiasdiener:
(I think we may need a new release after this PR 😬 )


inducer:
https://github.com/inducer/pyopencl/releases/tag/v2024.2.6 :crossed_fingers:

PR #770, opened 2024-06-19
Move more configs to pyproject.toml

alexfikl:
It's the future! :tada: 

This mostly removes `pytest.ini` and moves most of the things from `setup.cfg` here. 

Flake8 doesn't support `pyproject`, so it's not moved, but there is a project that hacks support in there if we want:
https://github.com/john-hen/Flake8-pyproject

### DIFF
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -10,4 +10,3 @@ dependencies:
 - ocl-icd
 - pocl
 - mako
-- pybind11
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,51 @@ sdist.exclude = [
     "setup.cfg",
 ]
 
+[tool.inducer-ci-support]
+disable-editable-pip-install = true
+enable-flake8-bugbear = true
+enable-flake8-isort = true
+
+[tool.flake8]
+max-line-length = 85
+inline-quotes = "double"
+docstring-quotes = "double"
+multiline-quotes = "double"
+select = [
+    "B",    # flake8-bugbear
+    "C",    # flake8-comprehensions
+    "D",    # pydocstyle
+    "E",    # pycodestyle
+    "F",    # pyflakes
+    "I",    # flake8-isort
+    "N",    # pep8-naming
+    "Q",    # flake8-quotes
+    "W",    # pycodestyle
+]
+extend-ignore = [
+    "E123", # closing bracket does not match indentation
+    "E126", # continuation line over-indented
+    "E127", # continuation line over-indented
+    "E128", # continuation line under-indented
+    "E226", # missing whitespace around arithmetic operator
+    "E241", # multiple spaces after comma
+    "E242", # tab after comma
+    "E265", # block comment should start with #
+    "E402", # module level import not at the top of file
+    "W503", # line break before a binary operator
+]
+exclude = [
+    "examples/gl_interop_demo.py",
+    "examples/gl_particle_animation.py",
+    "pyopencl/compyte",
+]
+per-file-ignores = [
+    "examples/pi-monte-carlo.py:N,B",
+    "examples/black-hole-accretion.py:N",
+    "examples/n-body.py:N",
+    "pyopencl/__init__.py:I001,I004,I005",
+]
+
 [tool.isort]
 line_length = 85
 lines_after_imports = 2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,17 @@ pocl = [
     "pocl-binary-distribution>=1.2",
 ]
 test = [
+    "flake8",
+    "flake8-bugbear",
+    "flake8-comprehensions",
+    "flake8-isort",
+    "Flake8-pyproject",
+    "flake8-quotes",
+    "isort",
     "mako",
+    "mypy",
+    "pep8-naming",
+    "pylint",
     "pytest>=7",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,6 +77,11 @@ combine_as_imports = true
 multi_line_output = 4
 extend_skip = ["pyopencl/__init__.py", "pyopencl/compyte"]
 
+[tool.pytest.ini_options]
+markers = [
+    "bitonic: tests involving bitonic sort"
+]
+
 [tool.mypy]
 warn_unused_ignores = true
 exclude = ["pyopencl/compyte"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,5 @@
 [build-system]
+build-backend = "scikit_build_core.build"
 requires = [
     "scikit-build-core >=0.9.3",
     "nanobind >=1.9.2",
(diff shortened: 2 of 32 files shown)


.test-conda-env-py3.yml
```diff
@@ -10,4 +10,4 @@ dependencies:
 - ocl-icd
 - pocl
 - mako
-- pybind11
+- nanobind
```
alexfikl: Is it a good idea to install `nanobind` from conda? I wasn't sure if it's better to just remove `pybind11` from here or replace it..

alexfikl:
The failures seem to be related to the release of numpy 2.0, but I'm not sure what it doesn't like since pyopencl should be built from source on the CI. Right?

EDIT: For reference, this fails on main too: https://github.com/inducer/pyopencl/actions/runs/9630426577/job/26560978801

matthiasdiener:
> Flake8 doesn't support `pyproject`, so it's not moved, but there is a project that hacks support in there if we want: https://github.com/john-hen/Flake8-pyproject

FWIW, I've been using flake8-pyproject in some of my projects, and it seems to work fine (see e.g. https://github.com/matthiasdiener/orderedsets/blob/551786a5b254372c097b3f676808d8e49dc0dcc8/pyproject.toml#L33)

.test-conda-env-py3.yml
matthiasdiener: AFAIK, the `nanobind` package itself does not need to be compiled (see https://github.com/conda-forge/nanobind-feedstock/blob/main/recipe/meta.yaml), so it's probably fine to use pypi and not conda for it.

alexfikl: Fair enough, I went ahead and just removed both `pybind11` and `nanobind` from the conda install.

alexfikl:
> FWIW, I've been using flake8-pyproject in some of my projects, and it seems to work fine (see e.g. [matthiasdiener/orderedsets@`551786a`/pyproject.toml#L33](https://github.com/matthiasdiener/orderedsets/blob/551786a5b254372c097b3f676808d8e49dc0dcc8/pyproject.toml#L33))

Yeah, I've used it too and it worked without any issues. 

One small problem with flake8 at the moment is also that the scripts in `ci-support` look for some strings in `setup.cfg` (e.g. `# enable-isort`), so that requires some more work..

inducer:
https://github.com/inducer/ci-support/commit/3cbd99af89b67ac2a616100a430168354329a19e should allow seamless use of flake8-pyproject.

pyproject.toml
```diff
@@ -28,41 +28,153 @@ classifiers = [
...
+    ".editorconfig",
+    "TODOs",
+    "run-*.sh",
+    "setup.cfg",
+]
+
+[tool.aksetup]
+disable-editable-pip-install = true
+enable-flake8-isort = true
+enable-flake8-bugbear = true
```
alexfikl: @inducer What do you think about something like this instead of the comments that were in the `setup.cfg` file? If nothing else, it makes them easy to parse going forward :grin:

pyproject.toml
```diff
@@ -28,41 +28,153 @@ classifiers = [
...
+    "B",    # flake8-bugbear
+    "C",    # flake8-comprehensions
+    "D",    # pydocstyle
+    "E",    # pycodestyle
+    "F",    # pyflakes
+    "I",    # flake8-isort
+    "N",    # pep8-naming
+    "Q",    # flake8-quotes
+    "W",    # pycodestyle
+]
```
alexfikl: I added this too, but I'm not very convinced, more for a small discussion. (I don't recall if you mentioned this at some point?)

I thought about it mainly for two reasons:
* removing the `enable-isort` (and others) grepping in `ci-support`: with this it can just install all the plugins it wants and the project will select the ones it wants.
* if we move to `ruff`, all these are bundled anyway and we'll need to select the ones we want.
* *it's unlikely at this point, but this would also guard against new added error codes.*

alexfikl:
With https://github.com/inducer/ci-support/pull/32 we could also remove `setup.cfg` completely. The only one that remains is pylint with the special `.pylintrc-local.yml` :confounded:

pyproject.toml
inducer: ~~That's obviously nicer, let's do that.~~ Maybe `inducer-ci-support` as the tool name? `aksetup` was the nickname for that hodgepodge of setuptools hacks that we're killing.

I couldn't help myself: here's a one-liner dotted-name-retriever for TOML:
```sh
python3 -c 'import toml, sys; data = toml.load(sys.argv[1]); [data := data[name] for name in sys.argv[2].split(".")]; print(data)' src/pyopencl/pyproject.toml project.readme
README.rst
```
Then we can get rid of the greppy stuff (mostly), while deprecating support for `setup.cfg`.

pyproject.toml
inducer: I think I prefer this, even. Getting ci-support to understand this won't be too much of a lift. I like it mostly because it allows us to avoid making up our own idiotic config bits.

pyproject.toml
alexfikl: That looks very nice! I'll go ahead and rename it.

pyproject.toml
alexfikl: > I like it mostly because it allows us to avoid making up our own idiotic config bits.

Yeah, that too! We can leave it in then, and slowly teach `ci-support` and everybody else to slowly migrate.

pyproject.toml
alexfikl: Done in https://github.com/inducer/pyopencl/compare/73ca584e3ab8186873d91d70d50a3f4cffcab6b5..e7b9adc6389d34d5cb9d72e848024456782ac7fc.

pyproject.toml
inducer: Now that I think of it: Why do we even need this? If we have the tools listed in `project.optional-dependencies.test`, we could just use that.

inducer: Hmm. I guess the "ruff" argument stands. :/ How close is ruff a this point? Should we just switch to it?

alexfikl: Hm, yeah, that's a good point. Sounds easier to standardize on some optional dependencies group name and just work with that if it exists!

I kind of like that better.

alexfikl: > Hmm. I guess the "ruff" argument stands. :/ How close is ruff a this point? Should we just switch to it?

It doesn't have a lot of the fancier pylint checks that actually execute the code (or whatever they do), but everything else is there and seems to be working very nicely!

I would say it's pretty solid at this point (since they got bored and re-implemented `black` as well :))

alexfikl: Here's a list of the pylint checks that are implemented
https://github.com/astral-sh/ruff/issues/970

inducer: Alright, let's say ruff is the future then, replacing flake8+plugins. Pylint can live for the moment.

alexfikl:
@inducer This is looking good to me and all the pieces from `ci-support` seem to be working nicely. Is there anything else missing? I would leave the transition to `ruff` for another PR.

inducer:
So I tried moving the whole thing to ruff, and I liked what I saw. :shrug: It's a bit of churn, but it's just really quite nice. I rather like the LSP integration. I imagine we'll do that (with time) across the whole stack.

inducer:
LMK how you feel about this.

inducer:
It's certainly much less of a mess than flake8.

alexfikl:
> So I tried moving the whole thing to ruff, and I liked what I saw. 🤷 It's a bit of churn, but it's just really quite nice. I rather like the LSP integration. I imagine we'll do that (with time) across the whole stack.

I'm all for it! It looks very nice and it seems to have caught some extra stuff :grin: 

> It's a bit of churn

Have you tried `ruff check --fix`? I mostly replaced isort with `ruff check --fix --select=I` and it works very well, not sure about the other rules..

PR #772, opened 2024-07-02
Add and fix some more ruff rules

alexfikl:
Was playing around with more rules to see what's in there.

This adds `flake8-logging-format`, `pyupgrade` and the builtin `ruff` rules to the mix. The `RUF` rules were the most annoying to fix, but most of them seemed reasonable. A lot of
```
[1.0] + my_list -> [1.0, *my_list]
```

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -30,7 +30,7 @@ jobs:
           python-version: '3.x'
 
       - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel==2.17.0
+        run: python -m pip install cibuildwheel==2.19.2
 
       - name: Build wheels
         shell: bash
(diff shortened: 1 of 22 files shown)


alexfikl:
This should be good now. The failure in cibuildwheel seems to be because some server is down:
```
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container
```

EDIT: Seems like that needs a bump in the action: https://github.com/pypa/cibuildwheel/releases/tag/v2.19.2

EDIT: Well, that fixed that issue, but now it's complaining about something else :(

alexfikl:
@inducer That seems to have fixed it! Thanks for the help :grin:

inducer:
Thx!

PR #774, opened 2024-07-15
Add example that profiles parallel sum

FattiMei:
# Description
This program profiles the parallel sum kernel when summing two arrays of increasing size, interpreting the data as char, int, float... the results are plotted with matplotlib, it's an external dependency but it's a common one.

# Rationale
When experimenting with GPU computing is useful to estimate an upper bound on performance, this PR offers an example of how you could use pyopencl events to profile a kernel (I suppose we are profiling only the execution time without the data transfers).
The results show a powerful idea in parallel computing: using shorter types improves throughput

# Possible corners
 * I assumed that the profiling times were in nanoseconds
 * I wanted to work with fp16, but not all devices support it so I commented it out. Maybe we could query at runtime if fp16 arithmetic is supported

### DIFF
--- a/examples/demo_flops.py
+++ b/examples/demo_flops.py
@@ -1,5 +1,6 @@
 import matplotlib.pyplot as plt
 import numpy as np
+
 import pyopencl as cl
 
 
@@ -10,6 +11,10 @@ src = """
         z[i] = x[i] + y[i];
     }
 """
+
+MAX_ALLOCATION_SIZE = 2 ** 30
+WARM_UP_RUNS = 4
+HOT_RUNS = 10
 
 
 # allocates buffers of increasing size, for each run do a parallel sum interpreting
@@ -33,7 +38,7 @@ if __name__ == "__main__":
         properties=cl.command_queue_properties.PROFILING_ENABLE
     )
 
-    buffer_size = [2 ** i for i in range(10, 31)]
+    buffer_size = [2 ** i for i in range(10, 31) if 2 ** i < MAX_ALLOCATION_SIZE]
     data = np.zeros((len(buffer_size), len(types)))
 
     for row, nbytes in enumerate(buffer_size):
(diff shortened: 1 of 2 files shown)


examples/demo_flops.py
```diff
@@ -0,0 +1,66 @@
...
+            header  = f'#define T {literal}\n'
+            kernel  = cl.Program(ctx, header + src).build().sum
+
+            event   = kernel(queue, (sums,), None, x, y, z)
+            event.wait()
+
+            FLOPS = 1e9 * sums / (event.profile.end - event.profile.start)
+            GFLOPS = FLOPS / 1e6
+
+            data[row, col] = GFLOPS
```
inducer: Arguably, this workload will be bandwidth-bound, so GB/s will be the more appropriate measure.

examples/demo_flops.py
```diff
@@ -0,0 +1,66 @@
...
+        x = cl.Buffer(ctx, cl.mem_flags.READ_ONLY,  nbytes)
+        y = cl.Buffer(ctx, cl.mem_flags.READ_ONLY,  nbytes)
+        z = cl.Buffer(ctx, cl.mem_flags.WRITE_ONLY, nbytes)
+
+        for col, (label, literal, sizeof) in enumerate(types):
+            sums    = nbytes // sizeof
+            header  = f'#define T {literal}\n'
+            kernel  = cl.Program(ctx, header + src).build().sum
+
+            event   = kernel(queue, (sums,), None, x, y, z)
```
inducer: You will probably want to repeat this a few times to cut down on measurement noise.

inducer: It's generally good practice to do a few "warmup" rounds before timing, to better measure the steady-state rate.

examples/demo_flops.py
inducer, line 1: Could you look over the CI failures?

inducer (commented):
Thanks for contributing this. Some comments below.

examples/demo_flops.py
FattiMei: This decision was made because it's common to evaluate gpu performance based on TFLOPS (and this number is computed with similar workloads) and especially highlights the fact that of course the flops go up when working with smaller types

examples/demo_flops.py
FattiMei: There are problem with caches however. In cpu runs I get crazy GFLOPS for medium size arrays because they already live in the cache, gpu doesn't seem to suffer from this problem.
But with the new commits one could decide to do no warmup runs and only one hot run so it's ok

PR #775, opened 2024-08-02
Generate nanobind type stubs

inducer:
This is an attempt to make use of [nanobind's stub generation](https://nanobind.readthedocs.io/en/latest/typing.html#stubs)

This does not work at all for now. Problems:

- [ ] How do we convince scikit-build-core to install `py.typed`?
- [ ] `nanobind.stubgen` produces a bunch of lines with `@property` that decorate nothing.
- [ ] Can editable installs work?
- [ ] `nanobind.stubgen` fails during build while trying to import `pytools` (because it's apparently trying to import `pyopencl`)


inducer:
Superseded by #833.

PR #780, opened 2024-08-26
choose_devices: don't short-circuit dev choice when answers available

matthiasdiener:
Otherwise, spurious errors can occur:

```console
# only single device available on this platform:

$ clinfo -l
Platform #0: AMD Accelerated Parallel Processing
 `-- Device #0: gfx90a:sramecc+:xnack-  
$ PYOPENCL_CTX=AMD:0 python -c "import pyopencl; pyopencl.choose_devices()"
  File "<string>", line 1, in <module>
  File "/p/lustre1/diener3/Work/enew/pyopencl/pyopencl/__init__.py", line 1674, in choose_devices
    raise RuntimeError("not all provided choices were used by "
pyopencl._cl.RuntimeError: not all provided choices were used by choose_devices. (left over: '0')
```

`PYOPENCL_TEST=AMD:0` works fine in this context.

See also #781.

_Please squash_

### DIFF
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -1649,6 +1649,7 @@ def choose_devices(interactive: Optional[bool] = None,
     if not devices:
         raise Error("no devices found")
     elif len(devices) == 1 and not answers:
+        print(f"Choosing only available device: {devices[0]}")
         pass
     else:
         if not answers:
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -1648,7 +1648,7 @@ def choose_devices(interactive: Optional[bool] = None,
 
     if not devices:
         raise Error("no devices found")
-    elif len(devices) == 1:
+    elif len(devices) == 1 and not answers:
         pass
     else:
         if not answers:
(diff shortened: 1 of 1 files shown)


pyopencl/__init__.py
```diff
@@ -1648,7 +1648,7 @@ def parse_device(choice):
 
     if not devices:
         raise Error("no devices found")
-    elif len(devices) == 1:
+    elif len(devices) == 1 and not answers:
```
matthiasdiener, line 1651: Out of curiosity, why does the device selection have a different behavior from platform selection here, when there is only 1 "item" available - the user does not get to even see which device was chosen automatically in the 1-device case? Perhaps we could just remove this `elif` alltogether?

matthiasdiener: Another option could be to just print the automatically chosen device in the `len == 1` case.

inducer: The idea of the `elif` was to keep the interactive process as efficient as possible.

inducer: > Another option could be to just print the automatically chosen device in the `len == 1` case.

I like that. Want to add it to this PR?

matthiasdiener: > > Another option could be to just print the automatically chosen device in the `len == 1` case.
> 
> I like that. Want to add it to this PR?

Sure, what do you think of [55f3e0c](https://github.com/inducer/pyopencl/pull/780/commits/55f3e0c2a65491275a55ea25c9d69da22f179271)?

inducer: LGTM.

PR #786, opened 2024-09-16
Add pickling support for arrays, buffers

matthiasdiener:
Based on the `meshmode.DOFArray` pickling support.

TODOs:
- [ ] support MemoryPools
- [ ] add to docs

_Please squash_

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -370,6 +370,7 @@ def queue_for_pickling(queue, alloc=None):
         yield None
     finally:
         _QUEUE_FOR_PICKLING_TLS.queue = None
+        _QUEUE_FOR_PICKLING_TLS.alloc = None
 
 # }}}
 
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -342,6 +342,37 @@ _BOOL_DTYPE = np.dtype(np.int8)
 _NOT_PRESENT = object()
 
 
+# {{{ pickling support
+
+import threading
+from contextlib import contextmanager
+
+
+_QUEUE_FOR_PICKLING_TLS = threading.local()
+
+
+@contextmanager
+def queue_for_pickling(queue):
+    r"""A context manager that, for the current thread, sets the command queue
+    to be used for pickling and unpickling :class:`Array`\ s to *queue*."""
+    try:
+        existing_pickle_queue = _QUEUE_FOR_PICKLING_TLS.queue
+    except AttributeError:
+        existing_pickle_queue = None
+
+    if existing_pickle_queue is not None:
+        raise RuntimeError("array_context_for_pickling should not be called "
+                "inside the context of its own invocation.")
+
+    _QUEUE_FOR_PICKLING_TLS.queue = queue
+    try:
+        yield None
+    finally:
+        _QUEUE_FOR_PICKLING_TLS.queue = None
+
+# }}}
+
+
 class Array:
     """A :class:`numpy.ndarray` work-alike that stores its data and performs
     its computations on the compute device. :attr:`shape` and :attr:`dtype` work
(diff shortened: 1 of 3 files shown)


pyopencl/array.py
```diff
@@ -342,6 +342,37 @@ class _copy_queue:  # noqa: N801
...
+def queue_for_pickling(queue):
+    r"""A context manager that, for the current thread, sets the command queue
+    to be used for pickling and unpickling :class:`Array`\ s to *queue*."""
+    try:
+        existing_pickle_queue = _QUEUE_FOR_PICKLING_TLS.queue
+    except AttributeError:
+        existing_pickle_queue = None
+
+    if existing_pickle_queue is not None:
+        raise RuntimeError("array_context_for_pickling should not be called "
```
```suggestion
        raise RuntimeError("queue_for_pickling should not be called "
```

pyopencl/array.py
```diff
@@ -705,6 +736,63 @@ def __init__(
...
+        self._flags = state["_flags"]
+        self.size = state["size"]
+        self.nbytes = state["nbytes"]
+        self.events = []
+
+        if self.allocator is None:
+            context = queue.context
+            self.base_data = cl.Buffer(context, cl.mem_flags.READ_WRITE, self.nbytes)
+        else:
+            self.base_data = self.allocator(self.nbytes)
```
inducer: It looks like this would allocate redundantly if there were multiple arrays using the same buffer.

inducer: We probably want to allow the user to supply an allocator via the context manager.

test/test_array.py
```diff
@@ -2393,6 +2393,26 @@ def test_xdg_cache_home(ctx_factory):
 # }}}
 
 
+# {{{ test pickling
+
+def test_array_pickling(ctx_factory):
```
inducer: Test that this works for SVM-backed arrays.

pyopencl/array.py
```diff
@@ -705,6 +736,63 @@ def __init__(
                          "than expected, potentially leading to crashes.",
                          InconsistentOpenCLQueueWarning, stacklevel=2)
 
+    # {{{ Pickling
+
+    def __getstate__(self):
```
inducer, line 635: I think it'd be useful if  this worked for subclasses (liked `TaggedCLArray`), too. Should be tested, too.

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

pyopencl/array.py
matthiasdiener: What do you think of fdb352556d16fc6a192d23d033e62495a7300774 ?

test/test_array.py
matthiasdiener: Done as part of [25ebc7b](https://github.com/inducer/pyopencl/pull/786/commits/25ebc7bbd51e331d6f9c41267458c9ae090a3c5b)

pyopencl/array.py
matthiasdiener: What do you think of [25ebc7b](https://github.com/inducer/pyopencl/pull/786/commits/25ebc7bbd51e331d6f9c41267458c9ae090a3c5b)?

matthiasdiener: Hmm, do you have a suggestion on how to handle this case?

inducer: `b = a[:]` will make another array using the same buffer. (Any slice really.)

A way around this would be to make buffers picklable instead.

matthiasdiener: Is [490bb94](https://github.com/inducer/pyopencl/pull/786/commits/490bb945a387e91ff3c59fdf012ee34aa48ee08f) going in the direction you had in mind? Should this be a separate PR?

inducer: Yes, it does. I'm just not clear on why we need two separate `queue_for_pickling` decorators.

matthiasdiener: > I'm just not clear on why we need two separate `queue_for_pickling` decorators.

60e71eeb935740205c4e4a2c1bb0687f4e71574b removes the duplicate.

matthiasdiener:
pooled allocations: look at using `__reduce__`

PR #787, opened 2024-10-02
Array: use np.result_type directly

matthiasdiener:
Similar to https://github.com/inducer/pytato/pull/540

_Please squash_

### DIFF
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -75,6 +75,9 @@ def _get_common_dtype(obj1, obj2, queue):
         raise ValueError("PyOpenCL array has no queue; call .with_queue() to "
                 "add one in order to be able to perform operations")
 
+    # Note: We are calling np.result_type with pyopencl arrays here.
+    # Luckily, np.result_type only looks at the dtype of input arrays as of
+    # numpy v2.1.
     result = np.result_type(obj1, obj2)
 
     if not has_double_support(queue.device):
--- a/pyopencl/array.py
+++ b/pyopencl/array.py
@@ -32,7 +32,7 @@ import builtins
 from dataclasses import dataclass
 from functools import reduce
 from numbers import Number
-from typing import Any, Dict, Hashable, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 from warnings import warn
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


matthiasdiener:
This probably needs a bit more testing, but what do you think of the idea @inducer?

pyopencl/array.py
```diff
@@ -81,88 +75,16 @@ def _get_common_dtype(obj1, obj2, queue):
...
-    else:
-        o1_dtype = getattr(obj1, "dtype", type(obj1))
-        o2_dtype = getattr(obj2, "dtype", type(obj2))
-
-        o1_is_integral = np.issubdtype(o1_dtype, np.integer)
-        o2_is_integral = np.issubdtype(o1_dtype, np.integer)
-
-        o1_key = obj1 if o1_is_integral and not o1_is_array else o1_dtype
-        o2_key = obj2 if o2_is_integral and not o2_is_array else o2_dtype
+    result = np.result_type(obj1, obj2)
```
inducer, line 81: Add a comment (like the one in pytato) that we're passing pyopencl arrays to numpy, and that it doesn't seem to mind. Maybe add an analogous test.

inducer (commented):
I'm generally in favor. The less of this duplicated nonsense logic we drag around, the better.

pyopencl/array.py
```diff
@@ -81,88 +75,19 @@ def _get_common_dtype(obj1, obj2, queue):
...
-        o2_dtype = getattr(obj2, "dtype", type(obj2))
-
-        o1_is_integral = np.issubdtype(o1_dtype, np.integer)
-        o2_is_integral = np.issubdtype(o1_dtype, np.integer)
-
-        o1_key = obj1 if o1_is_integral and not o1_is_array else o1_dtype
-        o2_key = obj2 if o2_is_integral and not o2_is_array else o2_dtype
+    # Note: We are calling np.result_type with pyopencl arrays here.
+    # Luckily, np.result_type only looks at the dtype of input arrays as of
+    # numpy v2.1.
```
alexfikl: This needs a bump in `pyproject.toml` to `numpy>=2.1` if true?

inducer: Hmm... then this would put this PR on ice until we can actually depend on numpy 2.1? numpy 2.1 requires Py3.10, and, at least for now, the compat target is at 3.8. 3.8 [hits EOL](https://devguide.python.org/versions/) this month, so maybe this would justify a compat bump to 3.10 (globally, across all packages)?

matthiasdiener: What this comment intends to say is "np.result_type only looks at the dtype of input arrays until at least numpy v2.1". I tested this with numpy 1.23, and it seems to work fine.

pyopencl/array.py
matthiasdiener: Added in [5f668e2](https://github.com/inducer/pyopencl/pull/787/commits/5f668e2a28d44fb2b515bc0ee214b7e5a1611929)

pyopencl/array.py
matthiasdiener: Clarified the comment a bit in 0c37b58

alexfikl: Ah, I see! Then, yeah, that sounds completely fine!

PR #788, opened 2024-10-03
debug ci

matthiasdiener:


.github/workflows/ci.yml
```diff
@@ -67,6 +67,7 @@ jobs:
         -   uses: actions/checkout@v4
         -   name: "Main Script"
             run: |
+                export PROJECT_INSTALL_FLAGS="-v"
```
matthiasdiener: @inducer: can we make `-v` the default in ci-support? This greatly helps when trying to debug failures that happen because the underlying build system changes in some way (like in this case).

matthiasdiener:
It seems that https://github.com/wjakob/nanobind/pull/701 broke building pyopencl.

Perhaps monkeypatching `__init__` (like in `_add_functionality` for `Kernel`) does not work anymore?

.github/workflows/ci.yml
inducer: Sure. Could you do a PR?

matthiasdiener: https://github.com/inducer/ci-support/pull/39

src/wrap_cl_part_2.cpp
```diff
@@ -32,6 +32,11 @@
 
 
 
+void dummy_fn() {}
+PyType_Slot slots[] = {
```
inducer: I think I'd prefer to just move the implementation of the constructors to C++, instead of this slots hackery.

PR #796, opened 2024-10-18
set NOMINSIZE, LTO

matthiasdiener:
This results in a ~10% speedup for a simple microbenchmark like:

```python
import pyopencl as cl
import time

queue = cl.CommandQueue(ctx)

prg = cl.Program(ctx, """
__kernel void sum()
{

}
""").build()


knl = prg.sum

start = time.time()

for _ in range(100000):
    knl(queue, (1,), None)

# Intentionally disabled, we are only interested in the enqueue time
# queue.finish()

print(time.time() - start)
```


inducer:
Thanks!

PR #815, opened 2025-02-12
MNT: stop using deprecated numpy macros

tacaswell:
These were deprecated in np1.7 and will be removed in np2.3 (currently removed on numpy main branch).

xref https://github.com/numpy/numpy/pull/28254/

Without these changes pyopencl fails to compile with numpy main.


inducer:
Thanks!

PR #818, opened 2025-03-09
Fix ruff warnings about unused variables

alexfikl:
Not sure why these weren't caught before, but it seems they fixed a bug :grin:


inducer:
Thanks!

Issue #830, opened 2025-05-27
Kernel launch overhead


99991:
**Describe the bug**

Launching a kernel via PyOpenCL is 20 times slower than launching the same kernel with the corresponding OpenCL functions via `ctypes`.

**To Reproduce**

I discovered this problem while trying to implement bitonic sort. Here is a demonstration which first invokes the kernel with PyOpenCL's call syntax and then switches to calling the same kernel directly via the OpenCL functions loaded from `libOpenCL.so` using `ctypes`.

Output with Intel Core i5-2400 CPU:

```
231.444 ms with PyOpenCL dispatch method
221.348 ms with PyOpenCL dispatch method
221.905 ms with PyOpenCL dispatch method
220.082 ms with PyOpenCL dispatch method
219.206 ms with PyOpenCL dispatch method
223.196 ms with PyOpenCL dispatch method
246.064 ms with PyOpenCL dispatch method
252.709 ms with PyOpenCL dispatch method
255.384 ms with PyOpenCL dispatch method
249.199 ms with PyOpenCL dispatch method
  8.495 ms with ctypes dispatch method
  8.350 ms with ctypes dispatch method
  8.540 ms with ctypes dispatch method
  7.898 ms with ctypes dispatch method
  7.979 ms with ctypes dispatch method
  7.343 ms with ctypes dispatch method
  7.300 ms with ctypes dispatch method
  9.271 ms with ctypes dispatch method
  7.459 ms with ctypes dispatch method
  7.111 ms with ctypes dispatch method
```

Output with NVIDIA V100 GPU:

```
 25.175 ms with PyOpenCL dispatch method
 24.560 ms with PyOpenCL dispatch method
 24.169 ms with PyOpenCL dispatch method
 23.987 ms with PyOpenCL dispatch method
 24.396 ms with PyOpenCL dispatch method
 23.612 ms with PyOpenCL dispatch method
 23.937 ms with PyOpenCL dispatch method
 24.184 ms with PyOpenCL dispatch method
 23.541 ms with PyOpenCL dispatch method
 23.887 ms with PyOpenCL dispatch method
  1.951 ms with ctypes dispatch method
  1.872 ms with ctypes dispatch method
  1.876 ms with ctypes dispatch method
  1.885 ms with ctypes dispatch method
  1.879 ms with ctypes dispatch method
  1.872 ms with ctypes dispatch method
  1.914 ms with ctypes dispatch method
  1.962 ms with ctypes dispatch method
[...]

PR #832, opened 2025-05-27
Warn about repeated Program.knl retrievals

### LINKED ISSUE #831: Warn about repeated kernel retrievals
It seems to be a performance gotcha that seems to be tripping people up. See #830 for an example.

inducer:
Closes #831.

### DIFF
--- a/pyopencl/__init__.py
+++ b/pyopencl/__init__.py
@@ -379,8 +379,14 @@ def enable_debugging(platform_or_context):
              stacklevel=2)
 
 
+class RepeatedKernelRetrieval(UserWarning):
+    pass
+
+
 class Program:
     def __init__(self, arg1, arg2=None, arg3=None):
+        self._knl_retrieval_count: dict[str, int] = {}
+
         if arg2 is None:
             # 1-argument form: program
             self._prg = arg1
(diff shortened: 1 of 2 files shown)


pyopencl/__init__.py
```diff
@@ -445,6 +451,17 @@ def __getattr__(self, attr):
             # but this will give an error if the kernel is invalid.
             knl.num_args  # noqa: B018
 
+            count = self._knl_retrieval_count[attr] = (
+                self._knl_retrieval_count.get(attr, 0) + 1)
+
+            if count > 1:
+                warn(f"Kernel '{attr}' has been retrieved more than once. "
```
matthiasdiener, line 460: These warnings will generally be deduplicated, but in light of https://github.com/inducer/pymbolic/pull/147, would it make sense just to issue the warning once (and maybe use a bool flag, not an int)?

inducer: That's a fair point, done.

fjarri:
Wouldn't it be easier to just cache the result?

inducer:
> Wouldn't it be easier to just cache the result?

Fair question, see https://github.com/inducer/pyopencl/issues/830#issuecomment-2913170648.

fjarri:
Ah, I see, I forgot that the kernel objects are mutable.

99991:
Would it be appropriate to ping a few people proactively who might encounter this warning? A simple regular expression already gives quite a few hits.

https://sourcegraph.com/search?q=context:global+pro%3Fg.*%3Fqueue.*%3Fglobal_size&patternType=regexp&sm=0

1. https://github.com/HenriquesLab/NanoPyx/blob/main/src/scripts/publication/Delay_GPU.py#L32
1. https://github.com/bh107/bohrium/blob/master/bridge/npbackend/bohrium/nobh/bincount.py#L213
1. https://github.com/coxlab/eyetracker/blob/master/coxlab_eyetracker/image_processing/cl_convolution.py#L584
1. https://github.com/mathuin/TopoMC/blob/master/bathy.py#L140
1. https://github.com/benma/pysph/blob/master/src/sph/radix_sort/radix_sort.py#L189
1. https://github.com/benma/pysph/blob/master/src/sph/sph.py#L328
1. https://github.com/enjalot/EnjaParticles/blob/master/teach/boids/boids.py#L58
1. https://github.com/enjalot/EnjaParticles/blob/master/teach/sph/clradix.py#L115
1. https://github.com/enjalot/adventures_in_opencl/blob/master/experiments/wave/wave.py#L89
1. https://github.com/silx-kit/pyFAI/blob/main/src/pyFAI/opencl/OCLFullSplit.py#L262
1. https://github.com/Probe-Particle/ppafm/blob/main/ppafm/ocl/field.py#L2341

inducer:
Thanks for highlighting this possibility! I've never done such a thing (both search for usage of a certain API pattern that I'm changing, and notified users). I don't know that I'm ready to blast out notifications. Since it's a non-breaking change, I think I will hold off.

PR #833, opened 2025-06-04
Typing

inducer:


matthiasdiener (commented):
Could probably also remove ./run-{mypy,pylint}.sh and related files from .gitignore here.

inducer:
> Could probably also remove ./run-{mypy,pylint}.sh and related files from .gitignore here.

Fair point, it's just clutter, and it's easy to revive from git.

PR #847, opened 2025-07-06
Fully (almost) type tools.py

alexfikl:
Inspired by all the new errors in #616, this adds types to pretty much everything in `tools.py` (on a bit of a best effort basis, since some of them may be too strict). Some small-ish parts are still missing:
* type checking for `get_gl_sharing_context_properties`, mostly due to missing the OpenGL bindings.
* `np.dtype` doesn't seem to like taking a dict (in `match_dtype_to_c_struct`), not sure why.
* `mako` still isn't typed and causing some small errors.
* Some issues around `cl.array.Array.data` not having the right type. This might just need some checking to make sure it's ok.

### DIFF
--- a/pyopencl/_monkeypatch.py
+++ b/pyopencl/_monkeypatch.py
@@ -27,9 +27,11 @@ from sys import intern
 from typing import (
     TYPE_CHECKING,
     Any,
+    Literal,
     TextIO,
     TypeVar,
     cast,
+    overload,
 )
 from warnings import warn
 
@@ -259,7 +261,42 @@ def kernel_set_arg_types(self: _cl.Kernel, arg_types):
                     devs=self.context.devices))
 
 
-def kernel_get_work_group_info(self: _cl.Kernel, param: int, device: _cl.Device):
+@overload
+def kernel_get_work_group_info(
+            self: _cl.Kernel,
+            param: Literal[
+                    _cl.kernel_work_group_info.WORK_GROUP_SIZE,
+                    _cl.kernel_work_group_info.PREFERRED_WORK_GROUP_SIZE_MULTIPLE,
+                    _cl.kernel_work_group_info.LOCAL_MEM_SIZE,
+                    _cl.kernel_work_group_info.PRIVATE_MEM_SIZE,
+                ],
+            device: _cl.Device
+        ) -> int: ...
+
+@overload
+def kernel_get_work_group_info(
+            self: _cl.Kernel,
+            param: Literal[
+                    _cl.kernel_work_group_info.COMPILE_WORK_GROUP_SIZE,
+                    _cl.kernel_work_group_info.GLOBAL_WORK_SIZE,
+                ],
+            device: _cl.Device
+        ) -> Sequence[int]: ...
+
+
+@overload
+def kernel_get_work_group_info(
+            self: _cl.Kernel,
+            param: int,
+            device: _cl.Device
+        ) -> object: ...
+
+
+def kernel_get_work_group_info(
+            self: _cl.Kernel,
+            param: int,
+            device: _cl.Device
+        ) -> object:
     try:
         wg_info_cache = self._wg_info_cache
     except AttributeError:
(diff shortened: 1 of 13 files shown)


pyopencl/tools.py
```diff
@@ -477,16 +482,16 @@ def _monkeypatch_svm_docstrings():
 
 # {{{ first-arg caches
 
-_first_arg_dependent_caches: list[dict[Hashable, object]] = []
+_first_arg_dependent_caches: list[Mapping[Hashable, object]] = []
```
alexfikl, line 486: This is mostly so that it's covariant in the values, not sure it's a good idea..

pyopencl/_monkeypatch.py
```diff
@@ -259,7 +259,7 @@ def kernel_set_arg_types(self: _cl.Kernel, arg_types):
                     devs=self.context.devices))
 
 
-def kernel_get_work_group_info(self: _cl.Kernel, param: int, device: _cl.Device):
+def kernel_get_work_group_info(self: _cl.Kernel, param: int, device: _cl.Device) -> int:
```
inducer: This isn't correct. For example `CL_KERNEL_COMPILE_WORK_GROUP_SIZE` returns a sequence of integers:

https://github.com/inducer/pyopencl/blob/b3bb32d6527bb7bae407004da9e7903066d3405e/src/wrap_cl.hpp#L5088-L5095

`object` is probably the right thing here.

pyopencl/tools.py
inducer: They're not modified through this list, so this is OK by me.

pyopencl/tools.py
```diff
@@ -1234,38 +1285,48 @@ def dtypes_match():
...
+        raise RuntimeError(
+            "end-of-line comments ('//') may not be used in code snippets")
 
     return code.replace("\n", " \\\n")
 
 
-class _SimpleTextTemplate:
-    def __init__(self, txt):
-        self.txt = txt
+class _TextTemplate(ABC):
```
inducer, line 1299: The subclasses of this could be dataclasses it seems.

pyopencl/_monkeypatch.py
alexfikl: Yeah, thought that would have been too easy :( Changed it to `object` and added some more checks and casts in https://github.com/inducer/pyopencl/compare/3d7d45e82efd6e92488dad748cf9781e27b32a00..bfc1af2939357dd2ef6ae4dbe45bfd4869f29412.

pyopencl/tools.py
alexfikl: Made them into dataclasses in https://github.com/inducer/pyopencl/compare/3d7d45e82efd6e92488dad748cf9781e27b32a00..bfc1af2939357dd2ef6ae4dbe45bfd4869f29412. The `mako` wrapper was a bit awkward, but it looks fine.

inducer:
Found a way that avoids all the casting.

Btw, given the choice between `cast` and `assert`, I strongly prefer `cast` for cases where we're just providing info to the type checker. If you're worried about overheads, I think it'd be relatively straightforward to statically remove all the casts in an import hook or so.

alexfikl:
> Found a way that avoids all the casting.

Nice! I thought of doing something like that too, but wasn't sure it would work :(

> Btw, given the choice between `cast` and `assert`, I strongly prefer `cast` for cases where we're just providing info to the type checker. 

Hm, I usually went by "if I need to cast, it probably means I didn't narrow the types enough for the type checker" and tried to avoid it. Yours is probably a better way to look at it when gradually adding types to things. :-?

inducer:
> Yours is probably a better way to look at it when gradually adding types to things. :-?

Not claiming it's a good way! And gradual typing mostly doesn't factor into it for the most part. Just trying to make sense of when to use what. Here's a different restatement:

- `assert` if the condition is a useful check to have at run time.
- `cast` for things the type checker needs to know but does not.

As a microbenchmark matter, the two have slightly different cost profiles, but I'm choosing to just regard both as "free", to avoid being pulled in spurious directions.

inducer:
Thx!

alexfikl:
> Just trying to make sense of when to use what.

I'll happily defer to your newly gained extensive experience in typing :grin:

> As a microbenchmark matter, the two have slightly different cost profiles, but I'm choosing to just regard both as "free", to avoid being pulled in spurious directions.

I can definitely agree with that!

PR #850, opened 2025-07-09
Add types to elementwise.py

alexfikl:
This adds types pretty much all things in `elementwise.py`. There are some remaining errors that I'm not sure how to best handle:

1. `context_dependent_memoize` expects something hashable as a first argument, mostly due to this:
https://github.com/inducer/pyopencl/blob/2acc9f7f1d40210a94bb665cdbe8aa3fc8a8ef22/pyopencl/tools.py#L583

However, the `cl.Context` that the functions in elementwise use is very much not hashable (?).

2. Added a debatable `ScalarLike` alias for some functions in `array.py`.
3. Deprecated `range` and `slice` in `ElementwiseKernel` and used named that don't shadow global functions instead. Went for this so that their types are not hidden in the `kwargs`.


inducer:
> However, the `cl.Context` that the functions in elementwise use is very much not hashable (?).

:thinking: What makes you say that? `Context` is supposed to be hashable.

https://github.com/inducer/pyopencl/blob/2acc9f7f1d40210a94bb665cdbe8aa3fc8a8ef22/pyopencl/_cl.pyi#L1004-L1005

pyopencl/typing.py
```diff
@@ -56,6 +56,7 @@
 
 Allocator: TypeAlias = "Callable[[int], _cl.MemoryObjectHolder | _cl.SVMPointer]"
 
+ScalarLike: TypeAlias = int | float | complex | np.number[Any] | NDArray[Any]
```
inducer:
```suggestion
ScalarLike: TypeAlias = int | float | complex | np.number[Any]
```
?

pyopencl/elementwise.py
```diff
@@ -280,16 +291,36 @@ def __call__(self, *args, **kwargs: Any) -> cl.Event:
...
-        slice_ = kwargs.pop("slice", None)
-        capture_as = kwargs.pop("capture_as", None)
-        queue: cl.CommandQueue | None = kwargs.pop("queue", None)
-        wait_for = kwargs.pop("wait_for", None)
+        if "range" in kwargs:
+            from warnings import warn
+
+            warn(f"Passing 'range' to '{type(self).__name__}.__call__' is "
+                 "deprecated and will not be allowed starting from 2027. Use "
+                 "'idx_range' instead.", DeprecationWarning, stacklevel=2)
```
inducer: Why introduce these `idx` versions?

alexfikl:
Hm, mostly an unwavering trust in pyright:
```
pyopencl/pyopencl/elementwise.py:483:2 - error: Argument of type "(context: Context, dtype: dtype[Any], idx_dtype: dtype[Any], vec_count: int = 1) -> Kernel" cannot be assigned to parameter "func" of type "(Hashable, **P@first_arg_dependent_memoize) -> RetT@first_arg_dependent_memoize" in function "first_arg_dependent_memoize"
Type "(context: Context, dtype: dtype[Any], idx_dtype: dtype[Any], vec_count: int = 1) -> Kernel" is not assignable to type "(Hashable, **P@first_arg_dependent_memoize)-> RetT@first_arg_dependent_memoize"
Parameter 1: type "Hashable" is incompatible with type "Context"
"Hashable" is not assignable to "Context" (reportArgumentType)
```

Not quite sure what it doesn't like then, since it should see the `__hash__` override. Any ideas?

inducer (commented):
Thx! Only two minor nits.

pyopencl/typing.py
alexfikl: Sure, I think that works for what I tried to use it in `array.py`. Maybe even make it local to that file?

I mostly added the `NDArray` part because a few parts in the code did `a = np.array(a)` and this would silence the type checker :grin:

pyopencl/elementwise.py
alexfikl: Wanted to make the keyword arguments explicit so that they get nice types and those are both shadowing global functions. Would you prefer to keep the `kwargs.pop()` from before?

inducer:
> Not quite sure what it doesn't like then, since it should see the `__hash__` override. Any ideas?

Note that this goes the other way! `"Hashable" is not assignable to "Context"`, not `"Context" is not assignable to "Hashable"`

pyopencl/elementwise.py
inducer: The shadowing only happens inside that function, I think. And that's not sufficiently grave (to me) to warrant a whole annoying deprecation cycle.

pyopencl/typing.py
alexfikl: Removed the `NDArray` part and moved the whole alias to `array.py`.

https://github.com/inducer/pyopencl/compare/ab7ae49dede8977292b4575c475be56feda6215b..f7c969d573655fb125812b64aa1d983cf288c996

pyopencl/elementwise.py
alexfikl: Yeah, fair enough. There was one usage of `slice` in the function, but it's no biggie. Removed the deprecations in
https://github.com/inducer/pyopencl/compare/ab7ae49dede8977292b4575c475be56feda6215b..f7c969d573655fb125812b64aa1d983cf288c996

alexfikl:
> Note that this goes the other way! `"Hashable" is not assignable to "Context"`, not `"Context" is not assignable to "Hashable"`

:facepalm: It makes a lot more sense when you actually read it, yeah :( Should be fixed now!

inducer:
Thx!

PR #864, opened 2025-10-04
Type cla.zeros_like.

kaushikcfd:


inducer:
Thanks! I don't think the Windows CL failure is related, see #865 for that.

PR #886, opened 2026-01-05
Build Linux aarch64 wheels

adeebshihadeh:
GitHub Actions [now offers](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) native arm64 runners. See all the runners currently offered [here](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

Verification:
* passed CI build [here](https://github.com/adeebshihadeh/pyopencl/actions/runs/20703593264/job/59429797260)
* installed the wheel from the above run in [openpilot CI](https://github.com/commaai/openpilot/pull/36984) (where we had been building the wheel ourselves previously), as an e2e sanity check

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -18,7 +18,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-15-intel, macos-14]
+        include:
+        - os: ubuntu-latest
+          cibw_archs_linux: x86_64
+        - os: ubuntu-24.04-arm
+          cibw_archs_linux: aarch64
+        - os: windows-latest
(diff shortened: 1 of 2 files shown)


inducer:
Looks great, thanks! Turns out Ruff 0.14.11 was released two minutes ago as of this writing, and *that* broke the Ruff CI here. :facepalm:

inducer:
I'm guessing you would like a release with that? I've just tagged 2026.1, which should land on the package index once [this build](https://github.com/inducer/pyopencl/actions/runs/20830295388) finishes.

adeebshihadeh:
Awesome, thanks!

PR #895, opened 2026-01-23
Add type annotations to algorithms

alexfikl:


pyopencl/algorithm.py
```diff
@@ -105,18 +125,20 @@ def copy_if(ary, predicate, extra_args=None, preamble="", queue=None, wait_for=N
 
     .. versionadded:: 2013.1
     """
+    # FIXME: other scan_dtypes below use uint?
     if len(ary) > np.iinfo(np.int32).max:
-        scan_dtype = np.int64
+        scan_dtype = np.dtype(np.int64)
     else:
-        scan_dtype = np.int32
+        scan_dtype = np.dtype(np.int32)
```
alexfikl: Not quite sure if this was on purposes, so wanted to bring it up?

pyopencl/algorithm.py
```diff
@@ -436,9 +463,27 @@ class RadixSort:
...
+    first_array_arg_idx: int
+
+    def __init__(self,
+                 context: cl.Context,
+                 arguments: str | Sequence[DtypedArgument],
+                 key_expr: str,
+                 sort_arg_names: Sequence[str],
+                 bits_at_a_time: int = 2,
+                 index_dtype: DTypeLike = np.int32,
+                 key_dtype: DTypeLike = np.uint32,
```
alexfikl, line 486: Are these dtypes always meant to be integers? I'm guessing yes because "radix sort". (need to update `key_dtype` attribute if yes)

pyopencl/algorithm.py
```diff
@@ -922,37 +1019,38 @@ def get_compress_kernel(self, index_dtype):
...
         kernel_name = self.name_prefix+"_count"
 
         from pyopencl.characterize import has_double_support
         src = _LIST_BUILDER_TEMPLATE.render(
                 is_count_stage=True,
                 kernel_name=kernel_name,
-                double_support=all(has_double_support(dev) for dev in
-                    self.context.devices),
+                double_support=all(
+                    has_double_support(dev) for dev in self.context.devices),
```
alexfikl: Should this be using `self.devices` instead of `self.context.devices`? Not quite sure what the use case is for setting `self.devices` to something different though..

pyopencl/scan.py
```diff
@@ -29,11 +29,12 @@
 from typing import TYPE_CHECKING, Any, cast
 
 import numpy as np
+from mako.template import Template
+from typing_extensions import override
 
 from pytools.persistent_dict import WriteOncePersistentDict
 
 import pyopencl as cl
-import pyopencl._mymako as mako
```
alexfikl, line 36: Was this on purpose? From a quick grep, this seems to be the only place that uses `_mymako`.

pyopencl/scan.py
```diff
@@ -1035,57 +1064,56 @@ def __init__(
...
+        self.context = ctx
+        self.dtype = dtype
 
         if neutral is None:
-            from warnings import warn
-            warn("not specifying 'neutral' is deprecated and will lead to "
-                    "wrong results if your scan is not in-place or your "
-                    "'output_statement' does something otherwise non-trivial",
-                    stacklevel=2)
+            raise ValueError("must provide a 'neutral' element in scan")
```
alexfikl, line 1075: Setting this to `None` would raise an exception below in `_process_code_for_macro(neutral)`, so it doesn't seem like it's supported at all?

pyopencl/algorithm.py
inducer: I used to be pretty sloppy about "dtype" vs "type". Not opposed to cleaning that up.

pyopencl/algorithm.py
inducer: Yep.

pyopencl/algorithm.py
inducer: `self.devices` is better.

pyopencl/scan.py
inducer: Sounds good, kill it.

pyopencl/scan.py
inducer: Yeah. Everywhere else should switch to `_mymako`, too. Mako isn't a hard dependency, and all `_mymako` does is provide a nicer error message. For type checking, you could go
```py
if TYPE_CHECKING:
    import mako
else:
    import pyopencl._mymako as mako
```

pyopencl/algorithm.py
alexfikl: Ah, I meant that this is using `np.int` and the other places are using `np.uint`?

inducer: Oh. For `copy_if`, I think `uint` will also be OK.

pyopencl/scan.py
alexfikl: Went for that all over the place: [e9cf1ff](https://github.com/inducer/pyopencl/pull/895/commits/e9cf1ffeb47c3f7e81d6b6d3988b872e90f4d513)

pyopencl/algorithm.py
alexfikl: Done in https://github.com/inducer/pyopencl/compare/ba5f3444c696cba824277fc3c8d4ece12a692081..5ed4558c800f84dde18514dd8e47577e8c3f89cd.

pyopencl/algorithm.py
alexfikl: Done in https://github.com/inducer/pyopencl/compare/ba5f3444c696cba824277fc3c8d4ece12a692081..5ed4558c800f84dde18514dd8e47577e8c3f89cd

inducer:
This is awesome, thanks! This is good progress towards cleaning up a largely-untyped, somewhat... organically grown mess.

PR #901, opened 2026-03-29
Fix new ruff errors

alexfikl:


alexfikl:
The wheel failure seems to be just a timeout of some sort.

PR #910, opened 2026-06-28
Update baseline

alexfikl:
This bumps `pythonVersion = 3.12` for pyright and sweeps all the new numpy errors under the rug :grin:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -199,6 +199,7 @@ MACOSX_DEPLOYMENT_TARGET = "10.14"
 skip = [
     "*-win32",
 ]
+repair-wheel-command = "delvewheel repair -w {dest_dir} --exclude ext-ms-win-dx-dxdbhelper-l1-1-0.dll {wheel}"
 test-command = ""
 before-all = "bash {package}/scripts/build-ocl-windows.sh"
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -199,6 +199,7 @@ MACOSX_DEPLOYMENT_TARGET = "10.14"
 skip = [
     "*-win32",
 ]
+repair-wheel-command = "delvewheel repair -w {dest_dir} --exclude ext-ms-win-dx-dxdbhelper-l1-1-0.dll {wheel}"
```
alexfikl, line 202: Not quite sure what that dll is 🤷 From what I can tell, the built wheels aren't tested? Should I add a little test like for macOS to check that this actually imports?

alexfikl: `dvelvewheel` seems to already have long lists of dlls it ignores and whatnot, so maybe this one is just too new (?): https://github.com/adang1345/delvewheel/blob/master/delvewheel/_dll_list.py

inducer: Thanks for sorting that out. I'd been meaning to take a look, but you beat me to it! :grin: 

I think I have reasonable confidence that this a system DLL that we shouldn't try to ship, so this seems just fine. I'm more worried about the develwheel command changing out from under us. At any rate, I'm happy to wait for folks to report that the wheels are broken, at which point they can (a) show they care and (b) help make a regression test.

inducer:
Thanks!

PR #924, opened 2026-09-17
Remove dead importlib dependency

alexfikl:
Missed this yesterday. We depend on Python 3.10+, so it's not needed anymore :grin:


inducer:
Thx!

